PR #9, opened 2025-04-04
scaffold

t-young31:
## Resolves #1

- Adds in the repo structure following the [standard go project layout](https://github.com/golang-standards/project-layout)
- Adds some opinionated pre-commit config (please feel free to suggest others)
- Adds in [dependabot](https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide) for docker/go/npm updates
- Adds a [GitHub security policy](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository)
- Includes Docker image with targets for `dev` and `release` (staging/prod) builds of the API and frontend, where the dev targets run hot-reloads
- Includes a top-level [makefile](https://github.com/ucl-arc-tre/portal/blob/e02076af40578c5a2052d3b9645cf307d052e8ad/Makefile) with useful commands to create the dev environment (`make dev`) and run the tests
- Skeletons out the README
- Adds a v0 [openapi schema](https://swagger.io/specification/) with codegen using [openapi-codegen](https://github.com/oapi-codegen/oapi-codegen) for the backend and [openapi-ts](https://heyapi.dev/openapi-ts/get-started) for the frontend
- Uses the default [vite+react scaffold](https://vite.dev/guide/#scaffolding-your-first-vite-project) with Typescript
- Adds some dev auth with oauth2-proxy for OIDC/SSO integration
- Adds some graceful connection/shutdown code for db and the go [gin](https://github.com/gin-gonic/gin) HTTP server
- Includes [gorm](https://github.com/go-gorm/gorm), which I hope we wont live to regret
- Adds workflows for
  + Lint, by running pre-commit
  + Test, unit level and e2e level (note the lack of Cypress because I didn't know what I was doing...). e2e test is very much a placeholder
  + Image build+push to ghcr on merge into main for the web frontend and api
  + Security with [gosec](https://github.com/securego/gosec) for static analysis

Suggestions very welcome, and sorry for the diff – scaffolding is always chaos 😅

PR #31, opened 2025-04-09
Autoformat frontend

jhughes982:
- This PR adds the prettier autoformatter to pre-commit,
- Prettier has also been added to `package.json` along with the relevant prettier config files,
- Happy to alter the specifics of the formatting config if needed,
- Also, if there is a standard approach to opening PR's and referencing issues for this project then let me know. I wasn't sure if it is preferred to open and reference an issue for every PR.

### DIFF
--- a/web/.prettierrc
+++ b/web/.prettierrc
@@ -1,0 +1,8 @@
+{
+  "trailingComma": "es5",
+  "useTabs": false,
+  "tabWidth": 2,
+  "semi": true,
+  "singleQuote": true,
+  "printWidth": 120
+}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,3 +49,10 @@ repos:
         entry: make web-lint
         pass_filenames: false
         always_run: true
+
+      - id: web-format
+        name: web-format
+        language: system
+        entry: make web-format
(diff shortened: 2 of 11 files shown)


web/.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
+  "trailingComma": "es5",
+  "useTabs": false,
+  "tabWidth": 2,
+  "semi": true,
+  "singleQuote": true,
```
t-young31, line 6: I'd prefer double for consistency with the rest of the codebase, but happy to be outvoted if you feel strongly

acholyn: I have no strong feelings so that's fine by me

jhughes982: Double quotes is fine by me 👍

t-young31:
> if there is a standard approach to opening PR's and referencing issues for this project then let me know. I wasn't sure if it is preferred to open and reference an issue for every PR.

If possible I'd rather reference and issue almost every time, just in case there's some discussion/additional useful context to be provided before implementation. given we discussed this in #9 I reckon this is a valid exception to that!

t-young31 (approved):
🚀

acholyn (approved):

Issue #12, opened 2025-04-07
Cypress e2e testing setup inc. accessibility, auth


t-young31:
**As a** developer
**I want** there to be [cypress](https://www.cypress.io/) end-to-end tests that run on PR into main
**So that** I can be confident of change I make is good to be deployed

## Acceptance criteria

- [ ] [maybe] Accessibility with [pa11y](https://github.com/pa11y/pa11y)
- [ ] Runs headless on `make test-e2e`

## Followups

- [ ] e2e authz tests

PR #32, opened 2025-04-10
init rbac

### LINKED ISSUE #13: RBAC authz scaffold

t-young31:
## Resolves #13 

- Adds in RBAC using [casbin](https://casbin.org/) to the API
- Seeds users and roles on create
- Enforces the RBAC in middleware with default deny
- Adds a `dev-psql` make target for getting a quick psql shell
- Converts the stub `/hello` endpoint into a `/profile` (as always, naming suggestions very welcome)
- Adds creating users that don't exist in the db in the `SetUser` middleware, and assigning them a default 'base' role
- Adds a cache of users using [hashicorp/golang-lru](https://github.com/hashicorp/golang-lru), which I think is safe, provided we add extra attributes to a user in a `UserAttributes` (naming suggestions welcomed!) model

> [!NOTE]
> We're going to need to add in a e2e test for the authz, as a followup to #12

### Checklist


internal/middleware/user.go
```diff
@@ -1,24 +1,81 @@
...
+	"github.com/hashicorp/golang-lru/v2/expirable"
+	"github.com/rs/zerolog/log"
+	"github.com/ucl-arc-tre/portal/internal/graceful"
+	"github.com/ucl-arc-tre/portal/internal/rbac"
 	"github.com/ucl-arc-tre/portal/internal/types"
+	"gorm.io/gorm"
 )
 
 const (
+	userCacheTTL      = 1 * time.Minute
```
t-young31, line 18: thoughts appreciated on this number. _probably_ safe to be longer

acholyn:
> We're going to need to add in a e2e test for the authz, as a followup to https://github.com/ucl-arc-tre/portal/issues/12
I thought we're using casbin instead of authz?

api.web.yaml
```diff
@@ -7,24 +7,29 @@ servers:
   - url: /api/v0
 
 paths:
-  /hello:
+  /me:
```
acholyn, line 10: `/profile` is probably more the norm

deploy/dev/docker-compose.yaml
```diff
@@ -8,6 +8,7 @@ services:
       DEBUG: true
       PORT: "8080"
       IS_DEV_DEPLOY: true
+      ADMIN_USERNAMES: "cceatyo-test@arctretest.onmicrosoft.com" # todo
```
acholyn, line 11: do we get accounts for this through the onboarding or..?

internal/middleware/user.go
acholyn: does this mean it refreshes the cache every minute?

t-young31: it means that any object in the cache can be there for a maximum of a minute

t-young31:
> I thought we're using casbin instead of authz?

ah yeah – using authz to mean authorization, rather than a specific library

internal/middleware/user.go
acholyn: ah, yes certainly longer but unsure how long exactly

api.web.yaml
```diff
@@ -7,24 +7,29 @@ servers:
...
           description: Unexpected error
 
 components:
   schemas:
-    HelloResponse:
+    MeResponse:
       type: object
       required:
-        - message
+        - username
```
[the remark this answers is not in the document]
t-young31: the username is what gets returned by the auth to identify the user, which shouldn't vary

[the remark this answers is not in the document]
acholyn: ah ok nice, you did mention having something for them to edit though, so maybe display name for that? (and this is what will be checked against the pdf certs)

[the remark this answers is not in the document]
t-young31: > maybe display name for that? (and this is what will be checked against the pdf certs)

yep 💯. will grab their name from EntraID if they're a UCL person, and let them modify it

deploy/dev/docker-compose.yaml
t-young31: 🤔 – setup from sample is entra id..(?) (happy to move this to slack)

acholyn, line 11:
```suggestion
      ADMIN_USERNAMES: "cceatyo-test@arctretest.onmicrosoft.com,ccaeaho@ucl.ac.uk," # todo
```

acholyn: @jhughes982 what's your ucl id code

t-young31:
```suggestion
      ADMIN_USERNAMES: "cceatyo-test@arctretest.onmicrosoft.com,ccaeaho@ucl.ac.uk,cceajeh@ucl.ac.uk"
```

gsvarovsky (approved):
♥

internal/middleware/authz.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (a *Authorizer) isAuthorized(ctx *gin.Context) (bool, error) {
+	userId := GetUser(ctx).ID.String()
+	resource := trimBaseURL(ctx.Request.URL.Path)
+	method := rbac.WriteAction
+	if ctx.Request.Method == "GET" || ctx.Request.Method == "HEAD" {
+		method = rbac.ReadAction
+	}
+	//roles, _ := a.enforcer.GetRolesForUser(userId)
+	//log.Debug().Any("roles", roles).Any("resource", resource).Any("userId", userId).Any("method", method).Msg("authz")
```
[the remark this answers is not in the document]
t-young31: deliberately left, but happy to rm if it makes you sad

cmd/web-api/main.go
```diff
@@ -1,19 +1,21 @@
...
 	handler "github.com/ucl-arc-tre/portal/internal/handler/web"
 	"github.com/ucl-arc-tre/portal/internal/middleware"
 	openapi "github.com/ucl-arc-tre/portal/internal/openapi/web"
+	"github.com/ucl-arc-tre/portal/internal/rbac"
 	"github.com/ucl-arc-tre/portal/internal/router"
 )
 
 func main() {
-	_ = graceful.NewDB()
+	rbac.Init()
```
jhughes982, line 14: Would it make sense to initialise the DB and RBAC separately, and pass in the DB to the RBAC setup?

t-young31: ah so the `NewDB` call before was just a stub to make sure the API could init the DB. now we have something that needs the db (the RBAC init) it's run in that

jhughes982: Makes sense, I was wondering whether it still makes sense to initialise them separately. No strong feelings about it, especially if the RBAC stuff and the DB are closely linked.

t-young31: fair. the RBAC & DB shouldn't be closely linked. I think there's an argument for separating the initialisations, but given the db init is idempotent I'm reasonably happy with `NewDB` initialising if it needs and handing over a connection. happy to change if you feel strongly

jhughes982: I would personally separate them. But I'll leave the final decision to you as the current implementation does make sense still.

internal/graceful/db.go
```diff
@@ -13,6 +14,16 @@ const (
 	initConnectRetryDelay = 1 * time.Second
 )
 
+// Initalise the database with the required types
+func InitDB() {
+	db := NewDB()
```
jhughes982, line 19: `NewDB` is called several times throughout the app initialisation, but after the DB has first been created the app connects to the same DB instance. Would a more descriptive function name make sense? At the moment every time I see `NewDB` being called it feels like we are about to create an entirely new DB instance.

Or am I mistaken in the way it works (or maybe this is just a Go convention)?

Again, I don't feel strongly about this. My questions are more to help me learn and understand this initial app structure and conventions.

t-young31: yeah – `new` is convention https://go.dev/doc/effective_go#composite_literals with the type of the thing returned

jhughes982 (approved):
Looks good to me. Works locally and all tests pass. I'll probably spend a bit more time looking into the finer details, but I didn't want to hold this up any longer.

PR #40, opened 2025-04-14
fix csp and middleware

t-young31:
use self in [default csp](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/default-src) and don't attach secure middleware to ping route


jhughes982 (approved):
Looks good, although is there a way for me to test this locally?

t-young31:
> is there a way for me to test this locally?

if our e2e used some self-signed certs yes, currently no

Issue #19, opened 2025-04-07
Agreeing to approved researcher agreement


r-roker-ucl:
## Description

**As a** user
**I want to** agree to the approved researcher agreement 
**So that** I can become and approved researcher 

## Acceptance criteria

- [ ] user can read a versioned version of the agreement and accept it

## Dependencies

- #48 required for e2e test

r-roker-ucl:
could pull text from GitHub

PR #44, opened 2025-04-15
add approved researcher agreement endpoint

### LINKED ISSUE #43: API endpoint to get the versioned approved researcher agreement
**As a** frontend dev
**I want** to be able to get the latest version of the approved researcher agreement
**So that** I can show it to the user so they can agree to it

t-young31:
## Resolves #43 

- Adds a `GET /agreements/approved-researcher` (preempting other agreement types) which returns some markdown and the id of it (so the user can agree to it)
- Adds an `agreements` service which handles the initialisation and retrieval of the latest approved researcher agreement from the db
- Copies in the approved researcher text from https://isms.arc.ucl.ac.uk/agreement_templates/rism14-approved_researcher_agreement. I thought about pulling dynmaically, but would need unnecessary github outbound connectivity + auth, which I don't think's worth it for something that's going to change _O_(yearly)
- Refactors both the db and rbac init to support more than just one object
- Does **not** add any tests, as this should be sufficiently covered by the e2e as a part of #19

### Checklist


internal/handler/web/profile.go
t-young31, line 1: just mv from main.go

jhughes982:
Is there a reason this needs to be on the backend? Could it not just be added as frontend content and displayed directly?

t-young31:
it needs to be versioned and we need the user id<->agreement id linkage

jhughes982 (approved):
Looks good to me.

PR #51, opened 2025-04-30
Adding in nextjs

### LINKED ISSUE #52: Switch frontend to NextJS
## Description

**As a** developer
**I want to** use NextJS
**So that** routing is easy

Context: https://github.com/ucl-arc-tre/portal/pull/39#discussion_r2062612306

acholyn:
### Checklist

- [x] added nextjs@14 to circumvent app routing error @ 15
- [x] updated nextjs template with our initial page
- [x] login works
- [x] dev version works
- [x] release version works
- [x] Documentation has been updated

---
### Questions/Comments/Concerns
- login was set up for react web app, not for statically served files - is there a new setting we change this to to fix this or do we have to return a certain html file to get around parent rerender behaviour?
- the release build fails since we no longer have an`index.html`, or an `app/dist`
- since nextjs is full stack would it make sense to look into seeing if it can take over for the gin sections? Might be able to leverage more dynamic updates


acholyn:
Maybe we direct to a new page instead of the same page? Would mean we're not really making best use of the react side of things tho
Remind me why we don't want to change the nginx config?

web/src/app/page.tsx
```diff
@@ -0,0 +1,65 @@
...
+    <div className={styles.page}>
+      <h1 id="title--portal">Welcome to the ARC Services Portal</h1>
+      <main aria-label="ARC portal main content">
+        {isAuthed === false && (
+          <div>
+            <Link href="/oauth2/start">
+              <button className="btn--login" id="login">
+                Login with UCL SSO
+              </button>
+            </Link>
```
t-young31, line 43:
```suggestion
            <button className="btn--login" id="login">
              <a href="/oauth2/start">Login with UCL SSO</a>
            </button>
```
think it's <a> for external links

acholyn: https://nextjs.org/docs/14/pages/api-reference/components/link

t-young31: `/oauth2/start` isn't a react route

web/next.config.mjs
```diff
@@ -0,0 +1,13 @@
+// @ts-check
+
+/** @type {import('next').NextConfig} */
+
+const nextConfig = {
+  output: "export",
+  // to work with eslint9?
+  eslint: {
+    ignoreDuringBuilds: true,
+  },
```
jhughes982, line 10: I think we do actually want eslint to run during builds.

Makefile
```diff
@@ -61,7 +61,7 @@ e2e-dependencies:
 	if [ ! -f nginx.key ]; then echo "creating tls certificates for nginx" && openssl req -x509 -nodes -days 365 -subj "/C=GB" -newkey rsa:2048 -keyout ./nginx.key -out ./nginx.crt; fi
 
 web-lint: ## Lint frontend web things
-	cd web && npm run lint
+	cd web && npm run lint && npm run lint:cypress
```
t-young31, line 64:
```suggestion
	cd web && npm run lint:all
```

web/eslint.config.mjs
```diff
@@ -0,0 +1,24 @@
...
+  allConfig: js.configs.all,
+});
+
+const patchedConfig = fixupConfigRules([...compat.extends("next/core-web-vitals")]);
+
+const config = [
+  ...patchedConfig,
+  ...ts.configs.recommended,
+  { ignores: [".next/*", "dist/*", "cypress/*", "cypress.config.ts"] },
+];
```
jhughes982, line 22: It looks like this is ignoring the cypress folder, but then you have another eslint config inside the cypress folder. Is this intentional?

web/cypress/eslint.config.mjs
t-young31, line 1: I'd be happy not linting this but 🤷🏼

web/eslint.config.mjs
acholyn: yes, I thought we still wanted to lint cypress but just ignore some rules

e2e/eslint.config.mjs
jhughes982, line 1: I assume this can be deleted?

web/package.json
```diff
@@ -1,38 +1,40 @@
...
   "scripts": {
-    "dev": "vite",
-    "build": "tsc -b && vite build",
+    "dev": "next dev --turbo",
+    "lint:all": "npm run lint && npm run lint:cypress",
+    "build": "next build",
+    "start": "next start",
     "lint": "eslint .",
-    "preview": "vite preview",
+    "lint:cypress": "eslint --config cypress/eslint.config.mjs cypress/",
```
jhughes982, line 11: Theres a few different `lint` commands here and it is perhaps a bit convoluted. The lint command is actually already built in with `npm run lint`.

In addition, the current `"lint": "eslint ."` command will try to run the linter across all files, including the built static files that are made during `npm run build`. I don't think this is the behaviour we want.

Ideally, the linter would run across all of our own code, and ignore anything else.

jhughes982: You may also not need the `lint: cypress` command if our configs are setup correctly. But I'm less sure about that.

acholyn: I followed this [article](https://blog.linotte.dev/eslint-9-next-js-935c2b6d0371) to get ESLint to work with next14 because they're not directly compatible - next15 is the one that officially works with eslint9 and eslint8 with next14 but eslint8 is no longer supported. We bypass the built in linting because it throws errors

acholyn: we have a specific command for linting cypress, which we could get rid of if we don't want to lint the contents of the cypress folder (and can remove the config for it)
the plain lint command should be used in the release build because there's no cypress reference and so shouldn't complain that cypress can't be found

jhughes982: At the moment, if you run a build locally with `npm run build` it won't pass the linter check because it is checking the built folders (try it locally). You may need to add them to the config to ignore them.

acholyn:
I also get that error locally, it shows my two account options in the screenshot

web/tsconfig.json
```diff
@@ -1,4 +1,27 @@
...
+    "strict": true,
+    "noEmit": true,
+    "esModuleInterop": true,
+    "module": "esnext",
+    "moduleResolution": "bundler",
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "jsx": "preserve",
+    "incremental": true,
+    "plugins": [
```
jhughes982, line 16: I'd suggest adding `"noUnusedLocals": true` here to prevent us from leaving unused variables around

jhughes982:
Local tests have been fixed now and I added a convenience cypress method for polling the backend API. This is useful if you end up running the frontend tests before the API is ready and you are left wondering why tests are not passing 😀

jhughes982 (approved):
Looks good to me.

PR #56, opened 2025-05-08
ignoring next 15

acholyn:
Dependabot suggests updating to next 15, but this will cause compatibility issues as there's a bug with the app router. Until this is fixed, we should ignore it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,3 +25,5 @@ updates:
     ignore:
       - dependency-name: "next"
         versions: ["^15"]
+      - dependency-name: "next"
+        versions: ["^19"]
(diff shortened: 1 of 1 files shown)


jhughes982:
Can you also ignore the update to react 19 as NextJs 14 is not officially supported by React 19 so we want to stick with React 18.

.github/dependabot.yml
```diff
@@ -22,3 +22,8 @@ updates:
       dependencies:
         patterns:
           - "*"
+    ignore:
+      - dependency-name: "next"
+        versions: ["^15"]
+      - dependency-name: "next"
+        versions: ["^19"]
```
jhughes982, line 29: Shouldnt this be `React`?

jhughes982 (approved):
Looks good

PR #64, opened 2025-05-12
added basic starter layout, pages, and components

### LINKED ISSUE #63: Create basic App layout
- To get things started, we should add a few basic pages to the portal so that we can kick-off conversations about the app flow.

## Definition of done
- [x] A basic layout is added to the NextJs setup with links to a few dummy pages
- [x] A page containing a basic form
- [x] Start a structure/template to follow for creating pages, page components, and lower-level UI components

jhughes982:
## Resolves #63 

- This PR adds a basic app layout so that we have a template to follow for creating different levels of components (i.e. pages, feature logic, UI etc.)
- A few dummy pages have been added, along with a basic form,
- No thought has been put into styling for now as this can be discussed separate to this code implementation,
- If anyone has any strong opinions, we can discuss and decide on the template structure to follow here.
- This PR is deliberately basic for now, and no logic has been hooked up to any backend, I wanted to keep each stage simple so we can discuss along the way.

I also made a few decisions in this PR that can dictate our style going forward. Let me know if you disagree with any of the following:
- Pages and components have their own associated `.css` file,
- Opted to not use css modules (it complicates the logic and I've worked on large codebases with plain css and it's worked fine),
- Pages have a central `page.tsx` entrypoint and feature components are linked within them (e.g. see the `approved-researcher` folder),
- Components that are specifically linked to a page can live inside a components folder at the same directory level as the `page.tsx` file.
- Components or logic that are shared across multiple files can live in the `components` folder at the `/src/components` location.

### Checklist


jhughes982:
Accessibility test is failing, I'll fix that after lunch unless someone has any time today?

web/src/app/approved-researcher/components/ApprovedResearcherForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+import "./ApprovedResearcherForm.css";
+
+export default function ApprovedResearcherForm() {
+  const [submitted, setSubmitted] = useState(false);
+  const [formData, setFormData] = useState({
+    name: "",
+    email: "",
+    userID: "",
+    department: "",
```
t-young31, line 13: don't think we need any of this 😄

acholyn: do we need their EntraID? and we said we wanted a display name they can control, but not sure if that's this form

t-young31: the backend already has their EntraID. agree the display name should probably be before this form – probably be prompted on first access?

jhughes982:
@acholyn the favicon needs to go in the `/app` directory. I favour suggesting changes to people's PR's rather than pushing directly to them (unless it's been agreed), otherwise we risk getting entangled in people making changes that may get out of sync.

web/src/app/approved-researcher/components/ApprovedResearcherForm.tsx
acholyn: so we probably only need one from the email and entra ID and this can be hidden from the user but can still be part of the form to link the doc to them. Unless the way we store the docs means they're already linked? (eg `docs/<entraID>/approved-researcher`)

jhughes982: Ah ok, I took the fields from [this thing called 'Approved Researcher form'](https://liveuclac.sharepoint.com/sites/ARCTrustedResearch-ProdSharePoint/_layouts/15/listforms.aspx?cid=YTc4YTdmMTctMjJlOC00OTAyLTg1NDktMmRmOWVkMmQ3ZGI0&nav=NjIyZmQzNGYtOGEyMC00YzdlLTgzYmMtYTkxM2MwNTRjZmRj) 😅

jhughes982:
> should still be in the /app directory, I only removed the next.svg and put in our own icon so the folder wouldn't disappear

Sounds good, still if you could suggest it as a change rather than a direct push that would be preferable 🙂

web/src/app/approved-researcher/components/ApprovedResearcherForm.tsx
jhughes982: As this PR is not really about completing the approved researcher process, I'd like to focus on that in a follow-up PR to keep this strictly about setting up a basic layout. Unless anyone objects?

web/src/lib/setupOpenApi.ts
```diff
@@ -0,0 +1,3 @@
+import { client } from "@/openapi/client.gen";
+
+client.setConfig({ baseUrl: "/api/v0" });
```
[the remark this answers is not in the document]
jhughes982: Nice! I'll remove it.

jhughes982:
The basics have all been set up here now. I'd like to iterate on this in follow-up PR's unless there are further refinements people would like here?

.vscode/settings.json
t-young31, line 1: I'd rather ignore vscode settings

jhughes982: Ah, sneaky vscode put that in automatically. I'll remove it.

web/src/app/profile/approved-researcher/page.tsx
```diff
@@ -0,0 +1,60 @@
...
+        } else {
+          setIsAuthed(false);
+        }
+      } catch (err) {
+        console.error("Profile fetch error:", err);
+        setIsAuthed(false);
+      }
+    };
+
+    checkAuth();
```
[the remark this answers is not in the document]
jhughes982: It's because we can't make a `useEffect` itself `async`. So in order to use async operations we must define an async function inside the `useEffect`.

t-young31 (approved):
🚀

PR #68, opened 2025-05-14
preventing content flashes

### LINKED ISSUE #67: Brief content flashes
- While the application is currently loading the auth state, it falls back to the default content which is typically the sign-in button.
- The fallback content might be a bit annoying if it briefly flashes on screen.
- There are 2 options, either show a loading state, or just block any content until everything is ready.

## Definition of done
- [x] Decide on the most appropriate fallback content while the app is loading (either a loading spinner or no content at all)
- [x] Implement the desired behaviour

jhughes982:
## Resolves #67 

- Just a quick follow-up to improve the UI behaviour while the auth state is being loaded.
- The updated solution in this PR would default to showing no content until the auth state is finished loading, an alternative would be to show a message such as `loading..` or a spinner (but I don't think @t-young31 prefers that option)
- More context provided in #67 so feel free to post opinions on what your preferred solution would be

### Checklist

### DIFF
--- a/web/src/app/UserTasks.tsx
+++ b/web/src/app/UserTasks.tsx
@@ -4,7 +4,9 @@ import { useAuth } from "./hooks/useAuth";
 import LoginFallback from "@/components/ui/LoginFallback";
 
 export default function UserTasks() {
-  const { isAuthed, userData } = useAuth();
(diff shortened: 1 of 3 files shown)


t-young31 (approved):
I like

PR #73, opened 2025-05-15
fix user set on create in middleware

### LINKED ISSUE #69: Accessing auth content sometimes needs a manual refresh
- On a completely fresh app start up (i.e. a fresh docker build and volumes) and going through the auth process for the first time, when redirected back to the app I get a 403 response from the api `profile` endpoint. Therefore, the app still shows the login button as if they have not signed in.
- But when I do a manual refresh, I get a 200 response and the auth content shows correctly.
- Is this a backend/database sync issue? Are the necessary permissions not set yet when the user is redirected for the first time?
- Or a cookie issue perhaps? Are the necessary credentials not set yet when the user is redirected back?
- I don't think this is a critical issue, but it would be nice to improve the user experience here

## Definition of done
- [ ] Understand what causes this behaviour
- [ ] Implement a fix if necessary

t-young31:
## Resolves #69 

🤦🏼. We'll catch this in a test with #74

### Checklist

### DIFF
--- a/internal/middleware/user.go
+++ b/internal/middleware/user.go
@@ -43,17 +43,17 @@ func (u *UserSetter) setUser(ctx *gin.Context) {
 	if username == "" {
 		panic("username header unset")
 	}
-	user, existsInCache := u.cache.Get(username)
-	if !existsInCache {
-		user, err := u.persistedUser(username)
-		if err != nil {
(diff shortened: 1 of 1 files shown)


jhughes982 (approved):
Tested locally and this does indeed fix #69 🎉

PR #75, opened 2025-05-16
approved researcher agreement e2e test

### LINKED ISSUE #74: e2e tests for agreeing to the approved researcher agreement

t-young31:
## Resolves #74

- Adds an end-to-end test for a user agreeing to the approved researcher agreement. Note: that this user is an admin, we'll need a followup at some point to add a non-admin user to the e2e environment, and test that they can agree too
- Fixes the go frontend to serve all the .html and .txt, recursively
- Adds a `GET /profile/agreements` route, to get a users the confirmed agreements, and modifies the the UI to not show the agreement and form if they've already agreed to it
- To make the test pass on multiple runs it's conditional on whether the user has already agreed to it

### Checklist


Dockerfile
```diff
@@ -65,10 +65,11 @@ ENTRYPOINT ["./web-api"]
 FROM scratch AS web-frontend-release
 
 COPY --from=builder /etc/passwd /etc/passwd
-COPY --from=builder --chmod=777 /app/web-frontend web-frontend
-COPY --from=web-frontend-builder /app/out .
+COPY --from=builder --chmod=777 /app/web-frontend /web-frontend
+COPY --from=web-frontend-builder /app/out /out
```
t-young31, line 69: changed as https://github.com/ucl-arc-tre/portal/blob/7818e8a154c52e1119206fcf7085b6c747f9965f/cmd/web-frontend/main.go#L21 now loops recursively and there are files which it doesn't have permission to view

api.web.yaml
```diff
@@ -76,3 +90,26 @@ components:
         agreement_id:
           type: string
           description: UUID of the agreement that has been agreed to
+
+    ConfirmedAgreement:
```
t-young31, line 94: naming suggestions welcomed!

web/src/app/profile/components/Profile.tsx
```diff
@@ -23,9 +23,7 @@ export default function Profile() {
           </p>
 
           <p>
-            <Link href="/profile/approved-researcher" className="btn--form">
```
t-young31, line 26: css didn't seem to exist(?)

internal/service/profile/main.go
```diff
@@ -27,3 +29,23 @@ func (s *Service) ConfirmAgreement(user types.User, agreementId uuid.UUID) error
...
 }
+
+func (s *Service) ConfirmedAgreements(user types.User) ([]openapi.ConfirmedAgreement, error) {
+	data := []struct {
+		Type      string
+		CreatedAt time.Time
+	}{}
+	result := s.db.Select("type", "user_agreement_confirmations.created_at").
+		Table("user_agreement_confirmations").
+		Joins("left join agreements on agreements.id = user_agreement_confirmations.agreement_id").
```
t-young31, line 40: if anyone can think of a better way to do this I'd be all ears

web/src/app/profile/approved-researcher/components/ApprovedResearcherAgreement.tsx
```diff
@@ -0,0 +1,59 @@
...
+  if (loading) return null;
+
+  if (!isAuthed) return <LoginFallback />;
+
+  if (agreementConfirmed)
+    return (
+      <div id="approved-researcher-agreement">
+        <p>Agreement confirmed ✔</p>
+      </div>
+    );
```
jhughes982, line 48: This logic is kind of repeated in the `AgreementForm.tsx` component, specifically this:

```
      {submitted && (
        <div className="submitted-box">
          <p>Agreement confirmed</p>
        </div>
      )}
```

That piece of code might be redundant since when your `setAgreementConfirmed` state is set the parent component will render `Agreement confirmed` anyway.

jhughes982 (commented):
Nice! A few comments so far.

web/src/app/profile/components/Profile.tsx
jhughes982: @acholyn I guess you will be cleaning up the css as part of your upcoming PR? There are currently some unused classes in a few different files.

web/cypress/e2e/app.cy.ts
```diff
@@ -21,4 +21,20 @@ describe("ARC Portal UI authenticated", () => {
     cy.contains("Loading...").should("not.exist");
     cy.contains("List of user tasks here").should("exist");
   });
+
+  it("admin user can agree to approved researcher agreement", () => {
+    cy.loginAsAdmin();
+    cy.waitForApi(); // poll until the backend API is healthy
```
jhughes982, line 27: You might be able to use a `beforeEach` to avoid repetition across tests.

web/src/app/profile/approved-researcher/page.tsx
```diff
@@ -9,9 +9,9 @@ export default function ApprovedResearcherPage() {
   return (
     <div className="profile-page">
       <h1>Become an Approved Researcher</h1>
-      <p>To gain access to advanced services, please complete the following agreement.</p>
```
jhughes982, line 12: Is the `className="profile-page"` still appropriate here? The same className is used in the actual `ProfilePage` component.

web/cypress/e2e/app.cy.ts
```diff
@@ -21,4 +21,20 @@ describe("ARC Portal UI authenticated", () => {
...
+    cy.visit("/profile/approved-researcher");
+
+    cy.get("#approved-researcher-agreement");
+    cy.get("body").then((body) => {
+      if (body.find("#approved-researcher-agreement-text").length > 0) {
+        cy.get("#approved-researcher-agreement-text").should("be.visible");
+        cy.get(".input-checkbox").check();
+        cy.get(".submit-button").click();
+      }
```
jhughes982, line 36: Is the `cy.get("#approved-researcher-agreement-text").should("be.visible");` necessary here? It looks like it should already be available via the first `if` check?

api.web.yaml
acholyn: maybe signed instead of confirmed? but I think this is also fine

Dockerfile
jhughes982: All still works locally for me 👍

web/cypress/e2e/app.cy.ts
acholyn: doesn't the first just check it's in the DOM and the `be.visible` check it's not being hidden by CSS or another elem?

web/src/app/profile/approved-researcher/page.tsx
acholyn: I could fix this in the renaming PR if preferred

web/src/app/profile/approved-researcher/components/AgreementForm.tsx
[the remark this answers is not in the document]
t-young31: I'm not sure if `AgreementCheckbox` implies the existence of a submit button?

> Also, there are a lot of left over css classNames that are no longer being used. These should be deleted to keep things clean and minimal.

👍🏼 thanks

web/src/app/profile/approved-researcher/components/AgreementText.tsx
```diff
@@ -9,7 +9,7 @@ type AgreementTextProps = {
 
 export default function AgreementText(props: AgreementTextProps) {
   return (
-    <section className="section">
+    <section className="section" id="approved-researcher-agreement-text">
       <Markdown>{props.text}</Markdown>
     </section>
```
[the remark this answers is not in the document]
t-young31: is used as an anchor for the cypress test

web/src/app/profile/approved-researcher/components/AgreementForm.tsx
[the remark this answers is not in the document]
acholyn: can confirm the design system has a checkbox

web/src/app/profile/approved-researcher/components/AgreementText.tsx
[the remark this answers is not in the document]
acholyn: true, but we could use section.section for that 🤷‍♀️ 
I'm not sure what the stance is on including selectors just for tests, @jhughes982 do you have any opinions? I don't know if it matters or not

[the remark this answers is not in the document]
t-young31: hmm, what happens when there are multiple `section`s?

web/src/app/profile/approved-researcher/page.tsx
t-young31: which doesn't exist – I'll rm!

web/src/app/profile/approved-researcher/components/AgreementText.tsx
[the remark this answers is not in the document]
jhughes982: If you open up the cypress UI, there is a tool that helps you identify the most appropriate selector based on your HTML. You just hover over your element and it gives you the code for the best selector to use.

web/src/app/profile/approved-researcher/components/AgreementForm.tsx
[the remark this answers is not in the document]
jhughes982: > I'm not sure if AgreementCheckbox implies the existence of a submit button?

Ah, I also just noticed it is actually wrapped in a form. So the name looks ok on reflection!

web/cypress/e2e/app.cy.ts
jhughes982: The `body` check is a bit overkill from my perspective, as every page is wrapped in a `body` as part of the default app layout. In addition, there are earlier checks such as `cy.get("#approved-researcher-agreement");` which would fail if the `body` was not available.

So you might be able to simplify this to just `get` the relevant selector and rely on the earlier checks passing as evidence that it's ok to proceed. I don't think theres any logic that would make the text hidden, so a simple `get` should be enough. Something like this will still work:

```
 it("admin user can agree to approved researcher agreement", () => {
    cy.loginAsAdmin();
    cy.waitForApi(); // poll until the backend API is healthy
    cy.visit("/profile/approved-researcher");

    cy.get("#approved-researcher-agreement");

    cy.get("#approved-researcher-agreement-text");
    cy.get(".input-checkbox").check();
    cy.get(".submit-button").click();

    cy.contains("Agreement confirmed").should("be.visible");
  });
```

These are small points, and I don't feel strongly. Just suggestions for keeping things brief.

jhughes982: The `should("be.visible")` can still be good practice though, so including that is a good idea too 👍

jhughes982: Yeah, I'd include `should("be.visible")` then, but maybe get rid of the `body` check. Up to you.

t-young31: the the `body` check needs to exist because the assertion on `#approved-researcher-agreement-text` being visible won't pass it if it's not

web/cypress/e2e/app.cy.ts
```diff
@@ -1,6 +1,7 @@
 beforeEach(() => {
   cy.clearCookies();
   cy.clearLocalStorage();
+  cy.waitForApi(); // poll until the backend API is healthy
```
jhughes982, line 4: Looks like it doesn't quite work as expected this way (its possible to spin up the containers and run the tests before the API is fully ready). Something like this gives all the needed functionality:

```
beforeEach(() => {
  cy.clearCookies();
  cy.clearLocalStorage();
});

describe("ARC Portal UI unauthenticated", () => {
  it("shows the ARC portal phrase on the homepage", () => {
    cy.visit("/");

    cy.get("#title--portal").should("be.visible");
    cy.contains("fibble").should("not.exist");
  });
});

describe("ARC Portal UI authenticated", () => {
  beforeEach(() => {
    cy.loginAsAdmin();
    cy.visit("/");
    cy.waitForApi(); // poll until the backend API is healthy
  });

  it("can be logged into as an admin", () => {
    cy.visit("/");

    cy.contains("Loading...").should("not.exist");
    cy.contains("List of user tasks here").should("exist");
  });

  it("admin user can agree to approved researcher agreement", () => {
    cy.visit("/profile/approved-researcher");

    cy.get("#approved-researcher-agreement");
    cy.get("body").then((body) => {
      if (body.find("#approved-researcher-agreement-text").length > 0) {
        cy.get("#approved-researcher-agreement-text").should("be.visible");
        cy.get(".input-checkbox").check();
        cy.get(".submit-button").click();
      }
    });
    cy.contains("Agreement confirmed").should("be.visible");
  });
});

```

web/cypress/e2e/app.cy.ts
jhughes982: Hmm, its passing for me locally without the body check. Happy for it to stay in if it makes the test more predicatble.

web/cypress/e2e/app.cy.ts
jhughes982: Yeah that is slightly strange, I'm not totally sure yet. I'll have to have a bit of a dig to understand that behaviour.

web/cypress/e2e/app.cy.ts
t-young31: oh weird. did you try without teardown between? i.e `make test-e2e-dev && make test-e2e-dev`

web/cypress/e2e/app.cy.ts
jhughes982: Don't let that hold up this PR though. Can submit a follow-up if needed.

jhughes982 (approved):
Looks good.

web/cypress/e2e/app.cy.ts
jhughes982: Tried all the different ways I can think of, including headless mode and via the cypress UI. They all pass for me.

t-young31: ohh hang on – your suggestion doesn't ensure that the text is visible(?)

jhughes982: I tested it with your original code and it still passes 🤷‍♂️

PR #79, opened 2025-05-16
moved waitForAPI command into login command

### LINKED ISSUE #78: Investigate API polling during cypress tests
- There seems to be a specific order of events needed for the API polling to work correctly during cypress tests

This works as expected and throws an error if the API is not ready yet:
```
    cy.loginAsAdmin();
    cy.visit("/");
    cy.waitForApi();
```

But this does not throw an `API unavailable` error and instead just fails the test directly:
```
beforeEach(() => {
  cy.clearCookies();
  cy.clearLocalStorage();
  cy.waitForApi(); // poll until the backend API is healthy
});

  it("can be logged into as an admin", () => {
    cy.loginAsAdmin();

    cy.visit("/");
    cy.contains("Loading...").should("not.exist");
    cy.contains("List of user tasks here").should("exist");
  });
```

Ideally, the `waitForAPI` command would be flexible enough so that it can be put anywhere. Although the command is there as a convenience anyway, so I don't think we should spend too long digging into this. If it is more hassel than it is worth, we can just remove it.

## Definition of done:
- [x] Find out why the `waitForAPI` command doesn't work as expected when placed before the `loginAsAdmin` command
- [x] Implement a fix so that the `waitForAPI` command can be used more flexibly
- [ ] If the fix turns out to be long winded, we can decide whether it is worth keeping or not (the fix was not long winded)

jhughes982:
## Resolves #78 

- This solution bakes the `waitForAPI` command into the `loginAsAdmin` command so that it will throw an error if it tries to do backend requests while the API is not ready.
- It's less modular, but prevents us from having to call the `waitForAPI` command in every test that needs backend interaction

### Checklist

### DIFF
--- a/web/cypress/e2e/app.cy.ts
+++ b/web/cypress/e2e/app.cy.ts
@@ -1,7 +1,6 @@
 beforeEach(() => {
   cy.clearCookies();
   cy.clearLocalStorage();
-  cy.waitForApi(); // poll until the backend API is healthy
 });
 
 describe("ARC Portal UI unauthenticated", () => {
--- a/web/cypress/support/commands.ts
+++ b/web/cypress/support/commands.ts
@@ -71,6 +71,6 @@ Cypress.Commands.add("loginAsAdmin", () => {
     log.snapshot("after");
     log.end();
 
-    cy.waitForApi(); // Wait for the API to be available
+    cy.waitForApi(); // poll until the backend API is available
   });
 });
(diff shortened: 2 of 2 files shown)


t-young31 (approved):
🚀

jhughes982:
Just before I merge this, I think there is another solution to this. We could potentially modify the nginx config so that the `ping` route does not sit behind the oauth2 authentication. That way I think you could simplify the use of the `waitForAPI` command so that it doesn't rely on the oauth2 flow. Or you could even possibly have the docker containers wait until the API is ready based on the status of the `ping` route and that way you wouldn't even need the `waitForAPI` command.

These are all optional alternatives, if you don't think it's worth our time we can just go with this solution.

t-young31:
I think this is the cleanest solution, so has my vote

PR #80, opened 2025-05-16
76/bem

### LINKED ISSUE #76: Updating to bem
## Description

**As a** developer
**I want to** use the [bem](https://getbem.com/naming/) naming convention in CSS
**So that** I use good practice and can better organise reusable CSS in a large codebase

## Acceptance criteria

- [ ] CSS classes should use bem where possible
- [ ] tests should be updated

acholyn:
### Checklist
- updates classnames to use bem convention
- some lightweight changes to make buttons more noticeable + agreement links wcag compliant
- updating tests to work with updated selectors

--
I think profile could probably do with a small restructure but will do that in a later pr


web/cypress/e2e/app.cy.ts
```diff
@@ -27,13 +27,9 @@ describe("ARC Portal UI authenticated", () => {
...
-    cy.get("body").then((body) => {
-      if (body.find("#approved-researcher-agreement-text").length > 0) {
-        cy.get("#approved-researcher-agreement-text").should("be.visible");
-        cy.get(".input-checkbox").check();
-        cy.get(".submit-button").click();
-      }
-    });
+    cy.get(".section--approved-researcher-agreement-text").should("be.visible");
+    cy.get("input[name='agreed'][type='checkbox']").check();
+    cy.get(".button--submit").click();
```
t-young31, line 32: this doesn't work on sequential tests. 

```bash
make dev
make test-e2e-dev && make test-e2e-dev
# ...
make: *** [test-e2e-dev] Error 1
```

web/src/app/profile/approved-researcher/components/AgreementForm.tsx
```diff
@@ -27,21 +27,20 @@ export default function ApprovedResearcherForm(props: ApprovedResearcherFormProp
...
             name="agreed"
             onChange={() => {
               setAgreed(!agreed);
             }}
             checked={agreed}
-            className="input-checkbox"
             required
-          />
-          <button type="submit" className="submit-button">
+          />{" "}
```
t-young31, line 41: ?

web/src/app/UserTasks.tsx
```diff
@@ -11,14 +11,15 @@ export default function UserTasks() {
...
         Username&nbsp;{userData!.username}. Roles:&nbsp;
         {userData!.roles.join(", ")}
       </p>
-
-      <div>Your tasks:</div>
-      <div>List of user tasks here (e.g. approved researcher process)</div>
+      <div className="tasks__wrapper">
+        <h3>Your tasks:</h3>
+        <div>List of user tasks here (e.g. approved researcher process)</div>
+      </div>{" "}
```
t-young31, line 22: ?

web/src/app/profile/approved-researcher/components/AgreementText.tsx
```diff
@@ -9,7 +9,7 @@ type AgreementTextProps = {
 
 export default function AgreementText(props: AgreementTextProps) {
   return (
-    <section className="section" id="approved-researcher-agreement-text">
+    <section className="section section--approved-researcher-agreement-text">
```
t-young31, line 12: css doesn't exist(?)

web/src/app/globals.css
```diff
@@ -1,6 +1,8 @@
 :root {
   --background: #ffffff;
   --foreground: #171717;
+
+  --cta-green: #165823;
```
jhughes982, line 5: Might be worth making this more specific. For instance, does this refer to the colour of the text or something else (e.g. border)?

web/src/app/home.css
```diff
@@ -0,0 +1,3 @@
+.title__wrapper {
+  justify-items: center;
+}
```
jhughes982, line 3: I don't think you need a separate `home.css` as the root page `/` is built in as `page.tsx` which already has an associated `page.css`. I'd use one or the other.

web/src/app/layout.tsx
```diff
@@ -14,9 +14,9 @@ export default function RootLayout({ children }: { children: React.ReactNode })
     <html lang="en">
       <body>
         <AuthProvider>
-          <div className="layout">
+          <div className="content__wrapper">
```
jhughes982, line 17: I don't think that's correct BEM, because your `content` class here is inside `content__wrapper`, it should be the other way around. Your `content` is the block and your `content__wrapper` is the element within the block. Unless I'm reading it wrong?

web/src/app/layout.css
```diff
@@ -1,10 +1,24 @@
...
-.main-content {
+main.content {
   flex: 1;
   padding: 2rem;
 }
+
+.page {
+  height: 100svh;
+  justify-items: center;
+}
```
jhughes982, line 15: It looks like the `layout.tsx` does not contain the `page` class?

web/src/app/not-found.tsx
```diff
@@ -1,10 +1,12 @@
 import Link from "next/link";
+import "./layout.css";
+import "./not-found.css";
```
jhughes982, line 3: Ideally, you wouldn't have to import `layout.css` in any file other than `layout.tsx` because every other page inherits the top level layout.

web/src/app/page.tsx
```diff
@@ -1,4 +1,5 @@
 import UserTasks from "./UserTasks";
+import "./home.css";
```
jhughes982, line 3: We have a `page.css` for that

web/src/app/page.tsx
```diff
@@ -7,10 +8,11 @@ export const metadata = {
 
 export default function Home() {
   return (
-    <div className="page">
-      <h1 id="title--portal">Welcome to the ARC Services Portal</h1>
-      <p>This portal allows UCL researchers to manage ARC services and tasks.</p>
-
+    <div className="page page--home">
```
jhughes982, line 11: The BEM structure is not entirely clear here, I was expecting to see a `page` block and a `title` block. Also, the `--home` could possibly be made more clear as it is not clear what the effects of the `--home` modification are. Ideally, your `--` modification naming should express what the modification does (e.g. centred, large etc.)

web/src/app/profile/approved-researcher/components/AgreementForm.css
```diff
@@ -1,32 +1,23 @@
...
   padding: 1rem;
 }
 
-.input-checkbox {
+input[name="agreed"][type="checkbox"] {
   margin: 0.5rem;
 }
 
-.submit-button {
+.button--submit {
```
jhughes982, line 11: I think `btn` is used elsewhere so we should aim to be consistent. I usually go for the longer form, but it's up to you.

web/src/app/profile/approved-researcher/components/AgreementForm.css
```diff
@@ -1,32 +1,23 @@
...
 
-.submitted-box {
-  margin-top: 1rem;
-  padding: 0.5rem;
-  background-color: #171717;
-  border-radius: 4px;
-  font-size: 0.9rem;
+  &:hover {
+    background-color: color-mix(in srgb, var(--cta-green), 60%, white 40%);
+  }
```
jhughes982, line 22:
```suggestion
  &:hover {
    background-color: color-mix(in srgb, var(--cta-green) 60%, white 40%);
  }
```

I think that is the official syntax?

web/src/app/profile/approved-researcher/components/AgreementForm.tsx
```diff
@@ -27,21 +27,20 @@ export default function ApprovedResearcherForm(props: ApprovedResearcherFormProp
...
+          I agree
+          <button type="submit" className="btn button--submit">
             Submit
```
jhughes982, line 44: Any reason for the mixing of `btn` and `button` here?

web/src/app/profile/approved-researcher/components/AgreementText.tsx
```diff
@@ -9,7 +9,7 @@ type AgreementTextProps = {
...
       <Markdown>{props.text}</Markdown>
     </section>
   );
```
jhughes982, line 15: Since the HTML already tells us it is a section, this could be simplified to make the BEM relationships work. Also, the `--` should represent visual modifications. You could try something like:

```
<section className="approved-researcher-agreement">
  <h2 className="approved-researcher-agreement__title">...</h2>
  <p className="approved-researcher-agreement__text">...</p>
</section>
```

jhughes982 (changes requested):
Thanks Amanda. A few comments so far.

web/src/app/profile/approved-researcher/components/AgreementForm.tsx
acholyn: linter thing, I'm used to seeing it and I think it just makes things more readable?

jhughes982: If you're referring to the `{" "}` I would remove them. If we end up with lots of them they start to pollute the code and look messy.

web/src/app/profile/approved-researcher/components/AgreementForm.css
t-young31: +1 for `button`

web/src/app/profile/approved-researcher/components/AgreementForm.tsx
acholyn: habitually used to `btn` as general class, which is your preference and I'll update them

web/src/app/layout.tsx
acholyn: I always read wrappers as the outside; when you have a candy, the candy is the content, the wrapper is the content wrapper

web/src/app/not-found.tsx
acholyn: Indeed, but it didn't seem like the not-found was picking it up, not sure why it didn't seem like it was inheriting. Will test again locally and see if I'm still getting the issue

web/src/app/layout.tsx
jhughes982: Yeah, but that's not how BEM works 😅

web/src/app/not-found.tsx
acholyn: yep seems to be fine now

web/src/app/layout.tsx
acholyn: hmmm, will have a think and I guess overhaul because content__content doesn't seem like a good choice

web/src/app/layout.css
jhughes982: It depends if you are using the `page` class across multiple components? If it is designed to be reusable then it could go in `globals.css`. But my gut feeling tells me it's not really necessary, at the moment the `.page` class feels more like a layout style. in which case it might be better combined with `.layout`

t-young31 (approved):
🚀

web/src/app/UserTasks.tsx
```diff
@@ -11,14 +11,15 @@ export default function UserTasks() {
   if (!isAuthed) return <LoginFallback />;
 
   return (
-    <div className="card">
+    <div className="card card--user-tasks">
```
jhughes982, line 14: it doesn't look like the styles have been defined anywhere for either `card` or `card--user-tasks`. I'd go through and delete any css that is not currently being used.

web/src/app/page.tsx
```diff
@@ -7,10 +8,11 @@ export const metadata = {
 
 export default function Home() {
   return (
-    <div className="page">
-      <h1 id="title--portal">Welcome to the ARC Services Portal</h1>
-      <p>This portal allows UCL researchers to manage ARC services and tasks.</p>
-
+    <div className="page__content__inner--centered">
+      <div className="title">
+        <h1 className="title__content">Welcome to the ARC Services Portal</h1>
```
jhughes982, line 13: I can't find a style definition for `title__content`, if classes are not being used they should be removed.

web/src/app/not-found.css
```diff
@@ -0,0 +1,3 @@
+.page__content__inner {
+  font-size: 1.2rem;
+}
```
jhughes982, line 3: So we need to be careful here. It looks like the css has already started to drift in a direction that will become difficult to manage as the application grows. For instance, the `not-found.tsx` file is currently utilising two css classes: `page__content__inner` and `page__content__inner--centered`. But only one of those classes is defined in the linked `not-found.css` file. 

It's much easier to track where things are defined if they are restricted to the scope of a component, then you can safely assume the utilised styles are contained in the import e.g. `./not-found.css`. Right now, I'd have to search the codebase to find out where `page__content__inner--centered` has been defined because it's not in the `not-found.css` file. This is fine for a small app, but will become unwieldy as the codebase gets bigger.

In addition, it looks like different pages are utilising variations of the `page...` css style, but it doesn't look like there is a central logical place where related `page` classes are defined. The styles are working for now because the classes are being inherited from the `layout.tsx`, but this could easily fall into a scenario where we expect a style to be loaded but it isn't (usually because it is not clear where the style has been defined).

The best way to keep things clean is to have dedicated styles per component by utilising reusable `.tsx` components. For instance, instead of reusing the `page` css class and all its variations across different files, you would instead create a reusable `.tsx` component and just use the css class once inside that component. That way, you can have confidence that all the classes being utilised in a component are defined in a central place that is easy to track.

I hope that makes sense? Happy to jump on a call and go through this if you like?

jhughes982 (commented):
I have further thoughts on the css strategy here. Apologies for holding this up, but I think it's worth discussing this now while the app is still in it's early stages and easy to adjust. Might be best to discuss ideas on a call.

web/src/app/not-found.css
acholyn: I think I get you; it would essentially be creating a `Page` component and then we wrap the content pages within that, so Home would be something like

```html
...
return (
<Page>
    <div className="title">
        <h1 className="title__content">Welcome to the ARC Services Portal</h1>
        <p>This portal allows UCL researchers to manage ARC services and tasks.</p>
      </div>
      <UserTasks />
</Page>

)
```

acholyn: would it then be better to extend the `page__content` in not-found.tsx to adjust the font size or add another class specifically for this

jhughes982 (approved):
To not hold up follow-on PR's, I think it's probably best to merge this in. Myself and @acholyn will have a catchup today to discuss the css strategy for cleanliness and maintainability going forward, so we can do a further PR with bits of cleanup there too.

PR #84, opened 2025-05-21
switch to pages router

### LINKED ISSUE #82: NextJS app -> pages router
## Description

**As a** developer
**I want to** use the NextJS pages router
**So that** I can use a strict CSP without having to serve the frontend dynamically with Node

## Context

Turns out the app router wasn't the right way to go with our static export due to the generated inline JS on export, which is blocked by the content security policy (CSP) https://github.com/ucl-arc-tre/portal/blob/4d58fcc998e9900250835def3a6ff3d794d0acdb/internal/middleware/secure.go#L24

The NextJS [suggested solution](https://nextjs.org/docs/app/guides/content-security-policy) is to run middleware that sets a nonce, which is incompatible with a static export.

Writeup: https://0xdbe.github.io/NextJS-CSP-AppRouter/

## Acceptance criteria

- [ ] No inline JS is generated on the production frontend build

## Dependencies

- [x] #80

t-young31:
## Resolves #82

- Switches from the NextJS app to pages router to allow strict CSP without serving the frontend dynamically
- Moves things around appropriately, so there are now `web/src/styles` and `web/src/hooks`. Components are all moved out into `web/src/components`
- Switches to using CSS modules as the builds breaks with `Global CSS cannot be imported from files other than your Custom <App>. Due to the Global nature of stylesheets, and to avoid conflicts, Please move all first-party global CSS imports to pages/_app.js. Or convert the import to Component-Level CSS`. I think this is a good idea anyway because we were seeing unintentional clashes, which are prevented by the class name mangling
 
### Checklist

- [x] Documentation has been updated


web/src/pages/profile/approved-researcher/index.tsx
```diff
@@ -0,0 +1,23 @@
+import ApprovedResearcherAgreement from "@/components/profile/approved-researcher/ApprovedResearcherAgreement";
+import Head from "next/head";
+
+export default function ApprovedResearcherPage() {
+  return (
+    <>
+      <Head>
```
acholyn, line 7: could we abstract this into a component that takes a `content` prop?

web/cypress/e2e/app.cy.ts
```diff
@@ -25,13 +25,14 @@ describe("ARC Portal UI authenticated", () => {
     cy.loginAsAdmin();
     cy.visit("/profile/approved-researcher");
 
-    cy.get("#approved-researcher-agreement");
+    cy.get("#approved-researcher-agreement"); // wait for load
 
     cy.get("body").then((body) => {
-      if (body.find(".approved-researcher-agreement__text").length > 0) {
-        cy.get(".approved-researcher-agreement__text").should("be.visible");
+      console.log(body.find("#approved-researcher-agreement-text"));
```
acholyn, line 31: is this for debugging or an intentional log? if the latter maybe we change it to console.info and include a message?

t-young31: debugging. cheers – will rm

acholyn (approved):
🚀

PR #86, opened 2025-05-22
85/set name

### LINKED ISSUE #85: Record users full name on first login
**As a** developer
**I want** to record the users full name on first access 
**So that** I can check it against the training certificate

## Folloups

- [ ] #89

acholyn:
### Checklist
- added new api route to post to `/profile`
  - new schema for `ProfileUpdate`
- added dialog which shows up if no chosen name set
- added service and handler for managing the update for `chosen_name`
- 


api.web.yaml
```diff
@@ -18,6 +18,13 @@ paths:
                 $ref: "#/components/schemas/ProfileResponse"
         default:
           description: Unexpected error
+    post:
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/ProfileResponse"
```
t-young31, line 27: I think this should be a new schema e.g. `ProfileUpdate`, then will also need to define the responses below this e.g. https://github.com/ucl-arc-tre/portal/blob/9024182498ed647cbb578499a5154809be65a14d/api.web.yaml#L14-L20

internal/service/profile/main.go
```diff
@@ -49,3 +49,10 @@ func (s *Service) ConfirmedAgreements(user types.User) ([]openapi.ConfirmedAgree
 	}
 	return agreements, result.Error
 }
+
+func (s *Service) SetUserChosenName(user *types.User) error {
+
+	result := s.db.Model(&types.User{}).Where("username = ?", user.Username).Update("chosen_name", user.ChosenName)
```
t-young31, line 55: probably want to select on the `id` rather than the username

internal/types/user.go
```diff
@@ -1,8 +1,10 @@
 package types
 
 type Username string
+type ChosenName string
 
 type User struct {
 	Model
 	Username Username `gorm:"uniqueIndex"`
+	ChosenName ChosenName `gorm:"index"`
```
t-young31, line 9: assume we don't need an index on this(?)

web/src/components/index/UserTasks.tsx
```diff
@@ -1,17 +1,55 @@
...
+    const dialogElement = dialogRef.current;
+
+    try {
+      const response = await fetch("/profile", {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify({ chosen_name: chosenName }),
+      });
```
t-young31, line 25: better to use the codegen-ed [function](https://github.com/ucl-arc-tre/portal/blob/d356730451e81800f7510fc1f449632a3ded310a/web/src/openapi/sdk.gen.ts#L31) for all the type safety goodness cf. https://github.com/ucl-arc-tre/portal/blob/9024182498ed647cbb578499a5154809be65a14d/web/src/hooks/useAuth.tsx#L29

t-young31 (commented):
Looking nice – as the [build error](https://github.com/ucl-arc-tre/portal/actions/runs/15186581213/job/42708360814#step:4:61) there needs to be a handler function called `PostProfile`, which calls the service

api.web.yaml
```diff
@@ -62,13 +81,27 @@ components:
...
           type: array
           items:
             type: string
+        chosen_name:
+          type: string
+
+    ProfileUpdate:
+      type: object
+      required:
+        - username
```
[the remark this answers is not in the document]
acholyn: ah ok nice, I wasn't sure if we'd need it to link it to the right user

web/src/components/index/UserTasks.tsx
```diff
@@ -37,7 +36,7 @@ export default function UserTasks() {
   return (
     <div>
       {!userData!.chosen_name && (
-        <dialog open ref={dialogRef} id={styles.chosenName}>
+        <dialog open ref={dialogRef} id={styles.chosenName} data-id="chosenName">
```
acholyn, line 39: have found my peeve with css modules now

acholyn:
I'm getting some odd test behaviour locally in that it keeps setting my name to the Test one, despite using my env variable for typing and checking, likely due to caching, I suspect

internal/handler/web/profile.go
```diff
@@ -20,8 +21,30 @@ func (h *Handler) GetProfile(ctx *gin.Context) {
...
+	user := middleware.GetUser(ctx)
+	update := openapi.ProfileUpdate{}
+	if err := ctx.ShouldBindJSON(&update); err != nil {
+		log.Err(err).Msg("Invalid JSON object")
+		ctx.Status(http.StatusNotAcceptable)
+		return
+	}
+	user.ChosenName = types.ChosenName(update.ChosenName)
+
+	if err := h.profile.SetUserChosenName(&user); err != nil {
```
t-young31, line 40: doesn't need to be a pointer
```suggestion
	if err := h.profile.SetUserChosenName(user); err != nil {
```
(then update the function signature)

internal/middleware/user.go
```diff
@@ -25,6 +25,10 @@ func GetUser(ctx *gin.Context) types.User {
 	return ctx.MustGet(userContextKey).(types.User)
 }
 
+func UserIsCacheable(user types.User) bool {
+	return user.ChosenName != "" && user.ChosenName != "Test Chosen Name"
```
t-young31, line 29: as mentioned on slack – I'll add a followup issue to not cache any user attributes

internal/middleware/user.go
```diff
@@ -43,17 +47,23 @@ func (u *UserSetter) setUser(ctx *gin.Context) {
 	if username == "" {
 		panic("username header unset")
 	}
-	if user, existsInCache := u.cache.Get(username); existsInCache {
-		ctx.Set(userContextKey, user)
-		return
-	}
```
t-young31, line 49: the cache get I think needs to remain before the db read...

web/cypress/e2e/app.cy.ts
```diff
@@ -38,3 +38,27 @@ describe("ARC Portal UI authenticated", () => {
...
+
+describe("Setting chosen name for user", () => {
+  const chosenName = (Cypress.env("chosenName") as string) || "Test Chosen Name";
+
+  it("prompts user to set chosen name", () => {
+    cy.loginAsAdmin();
+    cy.clearChosenName();
+    cy.visit("/");
+
+    cy.get("dialog[data-id='chosenName']").should("be.visible");
```
t-young31, line 50: I think we should be consistent with selectors. above uses `id`s. I'm not adverse to switching everything to `data-id` or `data-cy` (despite @jhughes982 's aversion 😉), but we should be consistent

web/src/components/index/UserTasks.module.css
```diff
@@ -0,0 +1,22 @@
+#chosenName {
```
t-young31, line 1: any reason this is has an id rather than being consistent?

web/src/components/index/UserTasks.tsx
```diff
@@ -1,17 +1,51 @@
...
+          <form onSubmit={handleSubmit}>
+            Please enter your name as you would choose to have it appear on forms related to our services.
+            <input type="text" name="chosenName"></input>
+            <Button type="submit">Submit</Button>
+          </form>
+        </dialog>
+      )}
       <p>
-        Username&nbsp;{userData!.username}. Roles:&nbsp;
+        Name: {userData!.chosen_name}. Username&nbsp;{userData!.username}. Roles:&nbsp;
```
t-young31, line 48: this doesn't get updated. perhaps should be `useState`?

t-young31 (commented):
nice!

web/cypress/e2e/app.cy.ts
acholyn: I reckon I'll change this one to `data-cy` because it's only there for cypress since CSS modules means I can't just select by the ID now (the aforementioned peeve)

web/src/components/index/UserTasks.module.css
t-young31: ah so no hypens! maybe just `.dialog` ?

web/cypress/e2e/app.cy.ts
t-young31: can we change all the `id`-s to `data-cy` not just this one?

internal/middleware/user.go
acholyn: confused because if I move the cache before persistedUser() then it complains that there's no user 😅

web/src/components/index/UserTasks.module.css
acholyn: wilco
feels very wrong but I know the modules makes this possible

acholyn: hopefully we won't be having any other dialogs on this page

internal/middleware/user.go
t-young31: I mean like move 
```go
	if user, existsInCache := u.cache.Get(username); existsInCache {
		ctx.Set(userContextKey, user)
		return
	}
```
back here, and leave
```go
	if UserIsCacheable(user) {
		_ = u.cache.Add(username, user)
	}
```
below

web/src/components/index/UserTasks.module.css
t-young31: I _think_ the name mangling means the only requirement is that there isn't another dialog in the component

web/cypress/e2e/app.cy.ts
acholyn: I'd argue it depends if the id is also being used for styling. in tests, yes let's stick to using data-cy for consistency

internal/middleware/user.go
t-young31: #89

internal/middleware/user.go
acholyn: ah, I was reading it as check if the user is cachable before checking in the cache for it because if it's not cachable, it won't be there anyway

web/src/components/index/UserTasks.tsx
acholyn: should I be updating the auth context for this or am I overthinking it

acholyn: Have added a state just for chosen name, was previously thinking for userData

t-young31 (approved):
🚀

internal/service/profile/main.go
```diff
@@ -49,3 +51,15 @@ func (s *Service) ConfirmedAgreements(user types.User) ([]openapi.ConfirmedAgree
 	}
 	return agreements, result.Error
 }
+
+func (s *Service) SetUserChosenName(user types.User) error {
+	const isValidPattern = `^[A-Za-z\s-]*(?:\p{L}\p{M}*)*$`
```
t-young31, line 56: any reason it's not `^[A-Za-z\s\-\p{L}\p{M}]*$` ?

PR #90, opened 2025-05-27
add mutable user attributes

### LINKED ISSUE #89: Remove caching of user attributes
**As a** developer
**I want to** not cache user mutable attributes
**So that** they are always correct

t-young31:
## Resolves #89

- Moves the `chosenName` attribute of a user into a `UserAttributes` object, allowing the safe caching of `User`
- Splits some of `/profile` out into `/auth`
- Drive by:
   + Add missed responses in API definition to `POST /profile/agreements`

### Checklist


web/src/components/index/UserTasks.tsx
```diff
@@ -30,7 +44,6 @@ export default function UserTasks() {
         if (!response.response.ok) {
           throw new Error(`HTTP error! status: ${response.response.status}`);
         } else {
-          console.log(response.response);
```
t-young31, line 33: assume this was for debug and can be safely removed @acholyn?

PR #106, opened 2025-06-02
assign approved researcher role

### LINKED ISSUE #105: Assign role when all approved researcher tasks complete
**As a** user
**I want to** be assigned the role of `approved-researcher` when I have completed the training and agreed to the approved researcher agreement
**So that** I can perform actions that require being an approved researcher

t-young31:
## Resolves #105

- Add a `updateApprovedResearcherStatus` method to the profile service to add the `approved-researcher` role to users that fulfil the conditions, which is called after an agreement conformation or training upload (so users can do both in whichever order)
- Adds some new enums to the openapi schema so we don't drift
- Updates the training upload e2e test to make sure the user is an approved researcher after they complete it
- Drive by:
  + removes the duplicated roles from the homepage (needs to move to `/profile` but all good for now)

### Checklist

### DIFF
--- a/internal/service/profile/roles.go
+++ b/internal/service/profile/roles.go
@@ -19,7 +19,7 @@ func (s *Service) updateApprovedResearcherStatus(user types.User) error {
 		log.Debug().Any("username", user.Username).Msg("Not yet agreed to approved resarcher agreement")
 		return nil
 	}
-	if hasTraining, err := s.HasNHSDTrainingRecord(user); err != nil {
+	if hasTraining, err := s.hasValidNHSDTrainingRecord(user); err != nil {
 		return err
 	} else if !hasTraining {
 		log.Debug().Any("username", user.Username).Msg("Not yet completed NHSD training")
--- a/api.web.yaml
+++ b/api.web.yaml
@@ -198,6 +198,7 @@ components:
       type: string
       enum:
         - approved-researcher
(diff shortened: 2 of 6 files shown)


internal/service/profile/roles.go
```diff
@@ -0,0 +1,34 @@
+package profile
+
+import (
+	"github.com/rs/zerolog/log"
+	"github.com/ucl-arc-tre/portal/internal/rbac"
+	"github.com/ucl-arc-tre/portal/internal/types"
+)
+
+func (s *Service) updateApprovedResearcherStatus(user types.User) error {
```
acholyn, line 9: Would it make sense to shift some of this onus to the front end? So if both the training cert and agreement have an `agreed===true` then we call the function to update?

t-young31: ah gotcha. it's just what's there to stop someone doing a manual request to `POST /profile/roles` and assigning themselves as an approved researcher without actually completing the agreement or training?

> Will it matter if they do both in quick succession?

I _think_ there's a potential race condition, but doubt It'd be hit. I think you'd need to submit the training and <few hundred ms later agree to the training

gsvarovsky: [I couldn't resist looking](https://gorm.io/docs/advanced_query.html#Locking)

PR #114, opened 2025-06-13
refactor into single users service

### LINKED ISSUE #113: Refactor profile and user service
## Description

**As a** dev
**I want to** minimise the coupling between the profile and user service
**So that** there is more modularity

t-young31:
### Checklist


api.web.yaml
```diff
@@ -275,12 +278,7 @@ components:
...
-        - people
-      properties:
-        people:
-          type: array
-          items:
-            $ref: "#/components/schemas/Person"
+    People:
+      type: array
+      items:
+        $ref: "#/components/schemas/Person"
```
t-young31, line 284: little simplification here because I'd expect that we'll return the same data for approved researchers who 'own' people, just not everyone

internal/service/users/people.go
t-young31, line 1: this is basically just a move of https://github.com/ucl-arc-tre/portal/blob/main/internal/service/people/main.go  – just GH can't deal with the diff

Issue #107, opened 2025-06-03
Create People Views


acholyn:
# Description
**As an** IAO/IAA
**I want to** see who is associated with my project | see all users in the TRE
**So that I** can manage their access | & edit their details (when needed)

# Acceptance criteria

- page for /people that displays, in a tabular fashion, the users.
  - [ ] for approved researcher: #109 
  - [x] for an admin: #108 
- [x] Go response object that differs based on role
- [x] get request for /people
- [x] post request for /people/update
- [x] service + handler

## Follow ups

- https://github.com/ucl-arc-tre/portal/pull/110#discussion_r2140462755
- https://github.com/ucl-arc-tre/portal/pull/110#discussion_r2140470443

acholyn:
closing as completed since #109 no longer planned

Issue #20, opened 2025-04-07
Visibility of data on completed researcher agreements and training


r-roker-ucl:
## Description

**As an** IG administrator 
**I want to** see management information including who has agreed to the approved research agreement, and completed their training 
**So that** I can effectively manage the service. 

## Acceptance criteria

- [x] IG have a role that lets them view/edit all training records
- [x] e2e test added for both with admin role

## Dependencies

- #2

t-young31:
suggestion: `/users` page

acholyn:
closed by #116 ?

PR #116, opened 2025-06-17
107/update people: post

acholyn:
related to #107 and #20, addressing the POST aspect

- adding the ability to add an NHS training record via user + date (if no date given, set to today) to allow back-dating
- tests to check this functionality

Drive by:
- accessibility colour fixes

image

### Checklist


jhughes982:
@acholyn are you still having issues with this PR? If so, I'd be happy to jump on a call today if needed?

acholyn:
test is failing in accordance with #133

jhughes982:
We might want to wait until #134 is merged in as that PR contains fixes for the test failures

api.web.yaml
```diff
@@ -138,6 +138,32 @@ paths:
         default:
           description: Unexpected error
 
+  /people/update:
```
t-young31, line 141: I think this should be `POST /people/{id}` the 'update'-ness is implied by the method. The `PersonUpdate` then doesn't need the `username` field

t-young31: oh sorry – I should have read the PR description 🤦🏼. I think, if we need to do it, we should do it later because it'll need a cache removal

acholyn: interesting, it hasn't seemed like that locally, but is that because locally cache operates differently?
It immediately booted me out when I updated my admin username and I had to destroy and rebuild it to get back in as an admin

acholyn:
> I think it should always be explicitly set, then a date-picker in the UI (sorry if it's already there!)

datepicker is there but I think that's easy to remove, but why would we want to remove an easy way to select a date? I think the input allows you to type it or to select from the datepicker and both are valid
could add a button saying "set to today" to make that aspect easier if we don't want to autoset it if it's blank

t-young31:
> datepicker is there

awesome 🙌🏼

> why would we want to remove an easy way to select a date

to encourage actually opening the PDF and putting in the right date!

api.web.yaml
acholyn: ahhh I see. Shall I just disable the edit button for now and we retain the rest of the functionality (though hidden) and follow it up in another PR?

t-young31: I'd remove it – I think there's some chance it'd get re-enabled (potentially by me!) without context. I'm also reasonably confident that we'll not need this functionality

acholyn:
image
🤷‍♀️

internal/service/users/training.go
```diff
@@ -19,7 +19,7 @@ func (s *Service) UpdateTraining(user types.User, data openapi.ProfileTrainingUp
 		return openapi.ProfileTrainingResponse{CertificateIsValid: ptr(false)}, nil
 	}
 	switch data.Kind {
-	case openapi.TrainingKindNhsd:
+	case openapi.TrainingKind("training_kind_nhsd"):
```
t-young31, line 22: should stay as the const

web/cypress/e2e/integration/people/admin-editing.cy.ts
```diff
@@ -0,0 +1,20 @@
+import { botBaseUsername } from "../../../support/commands";
+
```
t-young31, line 2: can we add a test here that the base role can't edit anyone?

internal/service/users/people.go
```diff
@@ -36,9 +44,70 @@ func (s *Service) GetAllPeople() (openapi.People, error) {
...
+
+		training_records = append(training_records, apiTrainingRecord)
+	}
+
+	return training_records, nil
+}
+
+func (s *Service) SetTrainingValidity(user types.User, trainingkind types.TrainingKind, date string) error {
+
+	confirmation_time, err := time.Parse(time.RFC3339, date)
```
t-young31, line 100: go: `confirmation_time` -> `confirmationTime`

internal/handler/web/people.go
```diff
@@ -31,3 +34,32 @@ func (h *Handler) GetPeople(ctx *gin.Context) {
...
+
+	// take the id from the url and get the person
+	person, err := h.users.GetPerson(id)
+	if err != nil {
+		setServerError(ctx, err, "Failed to get person")
+		return
+	}
+
+	if update.TrainingKind != nil {
+		date := *update.TrainingDate
```
t-young31, line 53: potential nil-pointer deref here. I'd make the field required in the API definition

t-young31:
@jhughes982 did you want to have a review of this?

web/src/components/people/AdminView.tsx
```diff
@@ -39,42 +44,93 @@ export default function AdminView() {
...
               </td>
+              <td>
+                <div className={styles.trainingRecord} data-cy="training">
+                  {person.training_record?.map((training: TrainingRecord) => (
+                    <div key={training.kind} className={styles.training}>
+                      {training.kind}
+                      {training.is_valid ? (
+                        <span className={styles.valid}>
+                          <CheckIcon />
+                        </span>
```
jhughes982, line 115: Not sure if this is a quirk of the test data or not, but I have two nhsd trainings in my user row `portal-jh-admin`. One of them is valid while the other is not, and it won't let me edit this a second time.

Any ideas?

Screenshot 2025-07-02 at 15 09 04

jhughes982 (changes requested):
Some initial comments to start while I keep reading through this.

web/src/components/people/AdminView.tsx
```diff
@@ -1,27 +1,32 @@
...
+import TrainingForm from "./TrainingForm";
 import ApprovedResearcherImport from "./ApprovedResearcherImport";
 
 const CheckIcon = dynamic(() => import("uikit-react-public").then((mod) => mod.Icon.Check), {
   ssr: false,
 });
-const XIcon = dynamic(() => import("uikit-react-public").then((mod) => mod.Icon.X), {
+export const XIcon = dynamic(() => import("uikit-react-public").then((mod) => mod.Icon.X), {
   ssr: false,
 });
```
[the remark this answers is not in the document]
acholyn: happy to put it in globals, since it's probably going to be used anytime we want to have a close button, I'd assume it would be used quite a lot, `components/globals.tsx`? or perhaps under the assets folder `components/assets/globals.tsx`?

web/src/components/people/AdminView.tsx
```diff
@@ -39,42 +44,93 @@ export default function AdminView() {
...
         <thead>
           <tr>
             <th>User</th>
             <th>Roles</th>
             <th>Agreements</th>
+            <th>Training</th>
           </tr>
         </thead>
-        <tbody>
+        <tbody className={styles.tbody}>
```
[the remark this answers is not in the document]
acholyn: ah, I've just had another look and it seems to have lost all my styling, probably in the merge updates so will redo this

web/src/components/people/AdminView.tsx
```diff
@@ -39,42 +44,93 @@ export default function AdminView() {
...
               </td>
-              <td className={styles.roles}>{person.roles}</td>
+              <td className={styles.roles}>{person.roles.join(", ")}</td>
               <td className={styles.agreements}>
                 {person.agreements.confirmed_agreements.map((agreement: ConfirmedAgreement) => (
                   <div key={agreement.agreement_type} className={styles.agreement}>
                     {agreement.agreement_type}
                     {agreement.confirmed_at ? (
-                      <span className={styles.confirmed}>
+                      <span className={styles.valid}>
```
jhughes982, line 95: I've forgotten exactly how this works. Is it possible to have an agreement entry in the DB for a user and it not be valid?

web/src/components/people/AdminView.tsx
```diff
@@ -39,42 +44,93 @@ export default function AdminView() {
...
                       </span>
                     )}
                     {agreement.confirmed_at && <small>{convertRFC3339ToDDMMYYYY(agreement.confirmed_at)}</small>}
                   </div>
                 ))}
               </td>
+              <td>
+                <div className={styles.trainingRecord} data-cy="training">
+                  {person.training_record?.map((training: TrainingRecord) => (
+                    <div key={training.kind} className={styles.training}>
```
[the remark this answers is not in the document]
acholyn: yep, this is the thing I mentioned before in one of our meetings and I wasn't sure how best to handle it. Having had a bit more time to think about it, we probably want to replace the one that's there if it's the same training type, but I think there was some discrepancy between whether `training_kind_nhsd` or `nhsd` is provided as the training kind and I didn't understand why

web/src/components/people/AdminView.tsx
acholyn: I think the disembodied edit is attached to that training row...I'm just going to have a rejig of this anyway

web/src/components/people/AdminView.tsx
```diff
@@ -39,42 +44,93 @@ export default function AdminView() {
     fetchPeople();
   }, []);
 
+  const updatePersonUI = (id: string, training?: TrainingRecord) => {
+    // get people object and find the person with the right id
```
jhughes982, line 48: How come the `TrainingRecord` is optional? What is the scenario where this would not be passed in?

web/src/components/people/TrainingForm.tsx
```diff
@@ -0,0 +1,136 @@
...
+      <form onSubmit={handleSubmit} noValidate className={styles.form}>
+        <p>Use this form to validate a training certificate. Make sure you check the date on the certificate.</p>
+
+        <select
+          id="training_kind"
+          name="training_kind"
+          value={trainingKind || ""}
+          onChange={(e) => {
+            setTrainingKind(e.target.value as TrainingKind);
+            if (errorMessage) setErrorMessage(null);
```
jhughes982, line 87: Am I reading this correctly? Is the error message being set to `null` if an error is found?

web/src/components/people/TrainingForm.tsx
```diff
@@ -0,0 +1,136 @@
...
+  ssr: false,
+});
+const AlertMessage = dynamic(() => import("uikit-react-public").then((mod) => mod.Alert.Message), {
+  ssr: false,
+});
+
+const TrainingKindOptions = {
+  //  is there a better way of doing this? Won't let me use type as a value
+  nhsd: "training_kind_nhsd",
+};
```
jhughes982, line 19: @t-young31 is there a backend implementation that fetches all training kinds? Is it worth building this functionality?

web/src/components/people/TrainingForm.tsx
acholyn: think I was trying to make it so if there's an error message and then the user is typing, clear the error

api.web.yaml
```diff
@@ -150,6 +150,32 @@ paths:
         default:
           description: Unexpected error
 
+  /people/{id}:
+    post:
+      parameters:
+        - in: path
+          name: id
+          required: true
```
jhughes982, line 158: Would it be worth adding a description here to tell us what the `id` represents?

web/src/components/people/TrainingForm.tsx
t-young31: not yet because there's only one training kind so far – I vote we add it when there is

internal/service/users/people.go
```diff
@@ -36,9 +44,70 @@ func (s *Service) GetAllPeople() (openapi.People, error) {
...
+	if err != nil {
+		return err
+	}
+	switch trainingkind {
+	case types.TrainingKindNHSD:
+
+		response := s.createNHSDTrainingRecord(user, confirmationTime)
+
+		return response
+	}
```
jhughes982, line 110: Whats the intended use of the switch here?

internal/service/users/people.go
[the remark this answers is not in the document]
t-young31: I've had a quick scooch through – would probably benefit from another set of eyes

web/src/components/people/AdminView.tsx
```diff
@@ -39,42 +42,91 @@ export default function AdminView() {
...
+    const personToUpdate = people!.find((person) => person.user.id === id);
+
+    if (personToUpdate && training) {
+      personToUpdate.training_record ??= [];
+      personToUpdate.training_record.push(training);
+    }
+  };
+
   if (!people) return null;
```
jhughes982, line 55: `if (!people) return null;`

This would be better as loading state while `people` are being fetched.

t-young31: I think impossible as there needs to be at least the user accessing that page in the db

web/src/components/people/AdminView.tsx
t-young31: currently there is no expiry on the agreement

internal/service/users/people.go
jhughes982: Since we don't have multiple types yet, I'd avoid adding code in anticipation. Keep it simple until you need to add extra logic. Think YAGNI as the default.

t-young31: given the training kind is part of the request I vote we keep the switch (so when/if other types are added you don't get the wrong behaviour), but have add a `default: panic("unsupported training kind")`

jhughes982: If its 2 for keeping it and 1 against then it looks like I'm outvoted 😁

internal/handler/web/people.go
```diff
@@ -33,6 +35,38 @@ func (h *Handler) GetPeople(ctx *gin.Context) {
...
+
+	default:
+		panic("unsupported training kind")
+
+	}
+
+	ctx.JSON(http.StatusOK, openapi.TrainingRecord{
+		Kind:        update.TrainingKind,
+		CompletedAt: &update.TrainingDate,
+		IsValid:     true,
```
t-young31, line 66: the validity should be determined by the `TrainingDate` (service probably needs a new method for that). happy to do it in a followup

web/src/components/assets/exports.tsx
```diff
@@ -0,0 +1,14 @@
...
+  ssr: false,
+});
+
+export const InfoIcon = dynamic(() => import("uikit-react-public").then((mod) => mod.Icon.Info), {
+  ssr: false,
+});
+
+export const TrainingKindOptions = {
+  //  is there a better way of doing this? Won't let me use type as a value
+  nhsd: "training_kind_nhsd",
```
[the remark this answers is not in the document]
jhughes982: As far as I am aware, you can't extract the literal value from the type (as the types won't be available at runtime).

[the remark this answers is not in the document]
jhughes982: Looking at the implementation though, I don't think there should be a need to redeclare the training kinds in `components/assets/exports.tsx`. You are currently using it like this:

```
{person.training_record.training_records?.map((training: TrainingRecord) => (
                    <div key={`${training.kind}-${training.completed_at}`} className={styles.training}>
                      {getHumanReadableTrainingKind(training.kind)}
                      {training.is_valid ? (
                        <span className={styles.valid}>
                          <CheckIcon />
                        </span>
                      ) : (
                        <span className={styles.invalid}>
                          <XIcon />
                        </span>
                      )}
                      {training.completed_at && <small>{convertRFC3339ToDDMMYYYY(training.completed_at)}</small>}
                    </div>
                  ))}
```

Which means a simple function to extract the final text of the string "training_kind_nhsd" should be all thats needed? Something like:

```
function extractTrainingKind(str) {
  const match = str.match(/kind_?(.*)$/);  
  return match ? match[1] : '';             
}

console.log(afterKind('training_kind_nhsd'));   // → "nhsd"

```

web/src/components/people/AdminView.tsx
```diff
@@ -1,80 +1,156 @@
...
+    );
+  }
   return (
     <>
+      {trainingDialogOpen && (
+        <TrainingForm id={id} setTrainingDialogOpen={setTrainingDialogOpen} updatePersonUI={updatePersonUI} />
+      )}
+      <p>
+        Please note the dates shown are when the agreement/training is <em>valid from</em>
+      </p>
```
jhughes982, line 102: I know this is picky, and just my own personal preference. But wouldn't it be more clear just to have the call say `Valid from: {date}`. Maybe it makes the cells too busy?

web/src/components/people/AdminView.tsx
```diff
@@ -1,80 +1,156 @@
...
+        <h2>Your Tasks</h2>
+        <Loading message="Loading users..." />
+      </div>
+    );
+  }
   return (
     <>
+      {trainingDialogOpen && (
+        <TrainingForm id={id} setTrainingDialogOpen={setTrainingDialogOpen} updatePersonUI={updatePersonUI} />
+      )}
```
jhughes982, line 99: You could simplify this so that the component gets rendered based on the `id` being set:

```
{selectedUserId && (
        <TrainingForm 
          id={selectedUserId} 
          setTrainingDialogOpen={() => setSelectedUserId(null)} 
          updatePersonUI={updatePersonUI} 
        />
      )}
```

That way, you wouldn't need the additional `setTrainingDialogOpen`.

Up to you if you want to implement it this way or not, I don't feel strongly about it.

web/src/styles/globals.css
```diff
@@ -2,16 +2,22 @@
   --background: #ffffff;
   --foreground: #171717;
 
+  --text--secondary: #595959;
+
   --green-300: #165823;
   --green-200: #237b35;
-  --green-100: #359749;
+  --green-100: #328543;
   --green-50: #64bb75;
```
jhughes982, line 10: What do the numbers represent here e.g. `50`

web/src/components/people/TrainingForm.tsx
```diff
@@ -0,0 +1,141 @@
...
+            name="display_date"
+            value={trainingDisplayDate}
+            onChange={(e) => {
+              updateTrainingDate(e.target.value);
+              setTrainingDisplayDate(e.target.value);
+              if (errorMessage) setErrorMessage(null); // Clear error as user types
+            }}
+            aria-invalid={!!errorMessage}
+            aria-describedby="trainingError"
+          />
```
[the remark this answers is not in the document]
t-young31: not that I'm aware of – should be fine without validation imo

web/src/components/people/AdminView.tsx
acholyn: I was thinking it would make the cells too busy, and I think if you've got it once clearly you don't need it 100 times everywhere

web/src/components/assets/exports.tsx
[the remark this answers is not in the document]
jhughes982: Ok, let's pick up on this when a new training kind is added 👍

PR #121, opened 2025-06-19
Updating cypress tests (adding fixtures and guards)

### LINKED ISSUE #115: Cypress testing bug?
## Description

**As a** developer
**I want to** understand why cypress is a bit iffy on the admin view tests
**So that** I have a nicer developer experience

When the people tests are moved to the bottom, errors occur because the content is incorrect, possibly due to the other tests where the user becomes an approved researcher, the user no longer is a base user in the people tests

see https://github.com/ucl-arc-tre/portal/actions/runs/15633489695/job/44043018840

jhughes982:
## Resolves #115 

- This PR does a bit of test refactoring to circumvent the tests failing because of modified database entries,
    - See the tagged issue for more discussion on this along with possible alternative solutions and additional thoughts,
- I also added an additional test for robusteness where we test the combination of being a 'base' role while also being an 'approved researcher',
- I also noticed an intermittent test failure as a result of local state not being ready when some of the tests were running. This caused some tests to fail around 40% of the time. As such, I have added an additional 'loading' guard to make sure the data is available by the time the tests actual test the assertion.

### Checklist

### DIFF
--- a/web/cypress/e2e/app.cy.ts
+++ b/web/cypress/e2e/app.cy.ts
@@ -6,9 +6,21 @@ beforeEach(() => {
 describe("People page content", () => {
   it("should show nothing to base role", () => {
     cy.loginAsBase();
+    cy.mockAuthAsBaseUser();
     cy.visit("/people");
+    cy.waitForMockedAuth();
 
     cy.contains("You do not have permission to view this page").should("be.visible");
+  });
+
+  it("should show content for base approved researcher", () => {
+    cy.loginAsBase();
+    cy.mockAuthAsBaseApprovedResearcher();
+    cy.visit("/people");
+    cy.waitForMockedAuth();
+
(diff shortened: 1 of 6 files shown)


web/cypress/e2e/app.cy.ts
```diff
@@ -6,11 +6,23 @@ beforeEach(() => {
...
+    cy.mockAuthAsBaseUser();
     cy.visit("/people");
+    cy.waitForMockedAuth();
 
     cy.contains("You do not have permission to view this page").should("be.visible");
   });
 
+  it("should show content for base approved researcher", () => {
+    cy.loginAsBase();
+    cy.mockAuthAsBaseApprovedResearcher();
```
acholyn, line 18: would it make sense to always mock when we do a login as base or admin? so just combine these two into one?

PR #123, opened 2025-06-19
updated auth hook loading variable name

### LINKED ISSUE #122: Refactor loading states
- At the moment we have a loading state that is triggered while the auth is being verified on the frontend. The state is simply called `loading`.
- The downside of this is that the name is a bit too generic, so if a local component also wants to implement loading state for other data, it has to pick a different name.
- I'd argue it's better to have a generic loading state name per component and have the auth state loading name be specific since it is shared across many other components.
- So I'd vote for calling the auth state `isAuthLoading` and the rest of the components can just use something like `isLoading` for their local API calls

jhughes982:
## Resolves #122 

- This PR refactors the auth state loading variable name to something that doesn't interfere with the loading name of individual components. Refer to the tagged issue for more discussion.
- I also removed a redundant line in a test as the text it was looking for would never exist.

### Checklist

### DIFF
--- a/web/cypress/e2e/app.cy.ts
+++ b/web/cypress/e2e/app.cy.ts
@@ -35,7 +35,6 @@ describe("ARC Portal UI authenticated", () => {
     cy.loginAsBase();
 
     cy.visit("/");
-    cy.contains("Loading...").should("not.exist");
     cy.contains("Username").should("exist");
   });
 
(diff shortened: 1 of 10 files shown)


acholyn (approved):
fair enough, this does make it more clear
🚀 🧠

PR #126, opened 2025-06-20
refactored accessibility tests

### LINKED ISSUE #125: Accessibility checks not working properly
- I noticed that the cypress accessibility tests are running too early in the page load lifecycle, before the main content has finished rendering. This is causing the tests to pass even though they are not testing the actual application content.

- Here is a breakdown of what is happening:
  1. **Async authentication check**: Pages make a `/api/v0/auth` request on load to determine user authentication status
  2. **Race condition**: Accessibility tests are running before this auth request completed
  3. So, the accessibility tests are only evaluating a minimal page state instead of the full rendered content

## Possible solution

- We can try adding a a generic wait condition to ensure the page layout is fully rendered before running accessibility checks. Something like this would force the tests to detect the content is ready before making the accessibility check:

```typescript
cy.visit(path);
cy.get("header").should("be.visible"); // Wait for page layout to load
cy.injectAxe();
...

```

- This would be the most generic solution as every page has a header.

## Definition of done
- [ ] The accessibility tests are updated to make sure they are actually testing the rendered content and not just a blank page.

jhughes982:
## Resolves #125 

- This PR refactors the accessibility tests to make them actually test our app content,
- Up until now, the accessibility tests have silently not been working as intended because the tests have been running based on blank content (see the tagged issue for more context)
- So this PR offers a solution to making sure the accessibility tests are running against our app content.
- I have also added additional accessibility blocks, so now our accessibility tests will run against three conditions: unauthenticated, authenticated with base role, authenticated with admin role

## Note
- Now that the tests are actually running against our app content it has found legitimate accessibility test failures. 
- I will try to fix these myself, but may have to catch up with whoever wrote the original code if I run into any issues.

### Checklist

### DIFF
--- a/web/cypress/e2e/accessibility.cy.ts
+++ b/web/cypress/e2e/accessibility.cy.ts
@@ -6,15 +6,53 @@ const pages = [
   { name: "Studies page", path: "/studies" },
   { name: "Assets page", path: "/assets" },
   { name: "Projects page", path: "/projects" },
+  { name: "People page", path: "/people" },
 ];
 
 describe("Accessibility - ARC Portal", () => {
-  pages.forEach(({ name, path }) => {
-    it(`should have no critical or serious accessibility violations on ${name}`, () => {
-      cy.visit(path);
-      cy.injectAxe();
-      cy.checkA11y(undefined, {
-        includedImpacts: ["critical", "serious"],
+  describe("Unauthenticated content", () => {
+    pages.forEach(({ name, path }) => {
+      it(`should have no critical accessibility violations on ${name} (login fallback)`, () => {
+        cy.visit(path);
+        cy.get("header").should("be.visible"); // wait for page layout to load
+        cy.injectAxe();
+        cy.checkA11y(undefined, {
+          includedImpacts: ["critical", "serious"],
+        });
+      });
+    });
+  });
+
+  describe("Base user content", () => {
+    beforeEach(() => {
+      cy.loginAsBase();
+    });
+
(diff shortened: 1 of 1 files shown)


web/src/styles/globals.css
```diff
@@ -2,8 +2,6 @@
   --background: #ffffff;
   --foreground: #171717;
 
-  --grey: color-mix(in srgb, var(--background) 70%, var(--foreground) 30%);
```
acholyn, line 5: is there a particular reason for removing this?

jhughes982: - Yes, it was causing a test failure as the contrast did not pass the accessibility requirement.
- I have ideas about the use of css variables, so I will open a separate ticket for that. In short though, using variable names like `grey` may not be too helpful as you may as well use the direct `grey` colour built into css. So defining css variables with names like `color-primary`, `color-seconday`, etc. might be better.

web/src/components/index/ChosenNameForm.tsx
```diff
@@ -57,7 +57,9 @@ export default function ChosenNameForm(props: ChosenNameFormProps) {
         <form onSubmit={handleSubmit} noValidate>
           <p>Please enter your name as you would choose to have it appear on forms related to our services.</p>
 
+          <label htmlFor="chosenNameInput">Chosen Name</label>
```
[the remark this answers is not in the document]
acholyn: yep, figures. Potentially if users find "chosen name" confusing then we should change it to "name"

PR #127, opened 2025-06-20
Cypress restructure

### LINKED ISSUE #124: Modularise cypress tests
- At the moment all the main cypress tests are in one big file (apart from the accessibility tests),
- It is starting to become more cumbersome to run the tests locally now as you have to wait for them all to finish,
- It would be good to split the tests up into a few different files so that test suites can be run in isolation,
- How exactly to split them up can be decided (e.g. by page, or feature etc.)

## Definition of done
- The cypress tests have been split into a few different files so that test blocks can be run in isolation

jhughes982:
## Note
- This PR can come after #126 is merged in.

- This PR proposes a new structure to the cypress tests as the single app.cy.ts was getting a bit too big,
- Have a look at the new cypress README.md which explains the new structure. But in short, I have separated things out into UI tests (which use mocked data) and integration tests (which have real database interactions).
- This is just one suggestion, so I am open to alternatives. Perhaps you can have a look and let me know if the structure makes sense or if it just adds more bloat.

### Checklist


web/src/components/nav/Nav.tsx
```diff
@@ -37,8 +37,8 @@ function NavItem({ href, icon, title }: { href: string; icon: ReactElement; titl
 export default function Nav() {
   return (
     <aside className={styles.sidebar}>
-      <nav>
-        <h4>Menu</h4>
+      <nav aria-label="Main navigation">
+        <h2>Menu</h2>
         <hr />
```
jhughes982, line 42: Changed due to the original version being flagged by the accessibility tests

acholyn (approved):

PR #134, opened 2025-06-25
new profile flow

### LINKED ISSUE #130: When to set chosen name
- I was testing going through the app as a new user and I have some thoughts about the flow of events, specifically around setting the users `chosen name`,
- At the moment, the user is prompted to set a chosen name as soon as they login for the first time. It seems like the app tries to force the user to set a chosen name by showing an alert message on the homepage that occupies the entire screen.
- However, I was able to bypass this by navigating to a different page in the URL bar. This then allows me to go through the approved researcher flow. But the issue is that when I try to upload a certificate it will give me an error message because no chosen name has been set.
- This has the potential to cause user confusion because the app doesn't prompt the user to go back to the homepage to set their chosen name.

## Suggested alternative flow
- The chosen name section is included as part of the approved researcher process instead of the homepage,
- e.g. the chosen name is listed as the first task in the approved researcher process, and they are not allowed to move forward (i.e. to upload a pdf or sign the agreement) until the chosen name is set.
- This way, the user is prompted through a very specific onboarding process with clear steps rather than different elements appearing in different places.
- We could build the UI such that it makes it very clear that this is a multi-step process. So, Step 1: Set your chosen name, Step 2: Upload your certificate, Step 3: agree to the form.
- We could grey out certain sections until the previous steps are complete.

jhughes982:
## Resolves #130 

- This PR proposes a new Profile flow (see tagged issue for more details),
- This is mostly just UI/UX changes, I think this new flow is more user intuitive and keeps related components together. But I'm open to more ideas,

## Automating light and dark mode tests
- Since I was getting so many test failures regarding light and dark mode, I decided to add both light and dark mode tests so that they always run together.

## Testing locally
- I'd recommend creating a clean slate to test this feature fully. So delete your current containers, remove your volumes, and start the containers up again. Then you can go through all the process steps from scratch.

### Checklist


internal/middleware/user.go
```diff
@@ -8,18 +8,27 @@ import (
...
 	"gorm.io/gorm"
 )
 
 const (
-	userCacheTTL      = 1 * time.Hour
 	userContextKey    = "user"
 	usernameHeaderKey = "X-Forwarded-Preferred-Username" // See: https://oauth2-proxy.github.io/oauth2-proxy/configuration/overview
 )
 
+func getUserCacheTTL() time.Duration {
```
t-young31, line 23: I’m not a fan of this - dev should, in my opinion, be as close to prod as possible. (sorry not looked at the rest of the PR) why by adding this is the user no-longer cacheable?

jhughes982: It's back to how it was now.

web/src/components/profile/ProfileStepProgress.module.css
```diff
@@ -0,0 +1,166 @@
...
+}
+
+.step-icon-pending {
+  background-color: #fef2f2;
+  color: #b91c1c;
+  border: 2px solid #fecaca;
+}
+
+.step-icon-current {
+  background-color: #0e8d3d;
```
acholyn, line 51: this might just be personal preference, but all green to me reads as everything is done, so maybe having current as red would make it more obvious it's active but not done?
onboard with pending as grey instead of red 👍

jhughes982: Yeah that could work, the ticks on the 'done' icons does differentiate it with 'current'. But Maybe a different colour could work. We could have green for 'done', amber for 'current', and red for 'pending'?

web/src/components/profile/Profile.tsx
```diff
@@ -1,35 +1,117 @@
...
+          <TrainingCertificate />
+        </div>
+      );
+    }
+
+    return (
+      <div className={styles["completed-message"]}>
+        <h2>Profile Complete!</h2>
+        <p>You have successfully completed all profile setup steps and are now an approved researcher.</p>
+      </div>
```
t-young31, line 99: I think this should be visible once a user finished all the steps (i.e. in addition to 'valid training')

t-young31 (commented):
first pass done – love it 🚀

t-young31:
I don't know whether this is new behaviour but when I navigate I get a flash of the `access restricted` component

web/src/components/profile/ProfileStepProgress.module.css
t-young31: +1 for grey instead of red

web/src/components/profile/Profile.tsx
```diff
@@ -1,35 +1,117 @@
...
+  }, [isAuthed]);
 
   if (authInProgress) return null;
 
   if (!isAuthed) return <LoginFallback />;
 
+  const isApprovedResearcher = userData?.roles?.includes("approved-researcher");
+  const hasChosenName = !!chosenName;
+
+  const steps: ProfileStep[] = [
```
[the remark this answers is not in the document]
jhughes982: Yeah, I imagine this page will be modified over time as we gather more requirements.

I suspect eventually we will have profile sub pages where the user can update preferences and see their profile data in a convenient way. Mapping this out on Figma would be good at some point.

web/src/components/profile/Profile.tsx
jhughes982: Yes, I was thinking that too. I'll have a look at implementing that.

jhughes982:
> I don't know whether this is new behaviour but when I navigate I get a flash of the `access restricted` component

I'll look into that.

web/src/components/index/UserTasks.tsx
```diff
@@ -23,25 +24,53 @@ export default function UserTasks() {
...
+        <p>
+          <strong>Chosen name:</strong> {chosenName}
+        </p>
+        <p>
+          <strong>Username:</strong> {userData.username}
+        </p>
+        <p>
+          <strong>Roles:</strong> {userData.roles.join(", ")}
+        </p>
+      </div>
```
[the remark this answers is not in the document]
t-young31: yeah – definitely on the profile page. I wonder what's helpful to see on the homepage. thoughts @acholyn ?

web/cypress/e2e/accessibility/profile.cy.ts
```diff
@@ -0,0 +1,158 @@
...
+
+        cy.visit("/profile");
+        cy.wait("@getProfile");
+        cy.wait("@getAgreements");
+
+        cy.contains("Set Your Chosen Name").should("be.visible");
+
+        cy.injectAxe();
+        cy.checkA11y(undefined, {
+          includedImpacts: ["critical", "serious"],
```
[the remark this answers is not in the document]
jhughes982: I made it into a reusable cypress command:

https://github.com/ucl-arc-tre/portal/pull/134/commits/2dde9d37f4674f48ea6bd28a61068102ec7c3537

web/src/components/profile/approved-researcher/TrainingCertificate.tsx
```diff
@@ -30,15 +32,30 @@ export default function TrainingCertificate() {
   const [isPDF, setIsPDF] = useState<boolean>(false);
   const [isSubmitting, setIsSubmitting] = useState<boolean>(false);
 
+  function base64Encode(file: File): Promise<string | null> {
```
t-young31, line 35: why nest?

web/src/components/profile/ProfileChosenName.tsx
```diff
@@ -38,26 +36,34 @@ export default function ChosenNameForm(props: ChosenNameFormProps) {
...
 
   return (
-    <>
-      <dialog open ref={dialogRef} className={styles.dialog} data-cy="chosenName">
-        <form onSubmit={handleSubmit} noValidate>
-          <p>Please enter your name as you would choose to have it appear on forms related to our services.</p>
+    <section className={styles.section} data-cy="chosen-name-form">
+      <h3 className={styles.title}>Set Your Chosen Name</h3>
+      <p className={styles.description}>
+        Please enter your name as you would choose to have it appear on forms related to our services.
```
finleybacon, line 59: Mentioned in the stand up earlier, but I suspect this wording could be improved to clarify that the name they enter is going to be matched/compared against the name on their submitted certificate. At the moment this is not quite explicit - and this prompt actually reads (to me at least) almost like a "what would you like us to call you" i.e. you need not give your full name.

web/src/components/profile/approved-researcher/TrainingCertificate.tsx
jhughes982: You mean inside the `TrainingCertificate` function? Theres no strict need I guess. Happy to move it outside the function 👍

web/src/components/profile/Profile.tsx
jhughes982: I added an backend method to pull in the training status. Can you have a look at that and let me know if I have covered everything required. Feedback on the implementation would be appreciated.

Now, as soon as the user has finished the last step, they see all the completed tasks with an updated message.

web/src/components/profile/ProfileChosenName.tsx
finleybacon: yep looks good!

api.web.yaml
```diff
@@ -251,6 +263,33 @@ components:
...
 
+    TrainingRecord:
+      type: object
+      required:
+        - kind
+        - is_valid
+      properties:
+        kind:
+          type: string
+          description: The type of training (e.g., "nhsd")
```
t-young31, line 274: this should probably be an enum rather than a bare string (and referenced in both `ProfileTrainingUpdate` and here), but very happy to do that in a followup

t-young31 (approved):
🙌🏼

PR #139, opened 2025-06-27
use TrainingRecordKind enum

### LINKED ISSUE #138: Use enum for training kind
- At the moment we have this in `api.web.yaml`:

```
TrainingRecord:
      type: object
      required:
        - kind
        - is_valid
      properties:
        kind:
          type: string // this should be an enum
          description: The type of training (e.g., "nhsd")
        is_valid:
          type: boolean
          description: Whether this training certification is currently valid
        completed_at:
          type: string
          description: Time in RFC3339 format when the training was completed
```

- @t-young31 pointed out that it would be better to have the type as an enum which will account for different kinds of training records in the future.

## Definition of done
- [ ] The `TrainingRecord` type is made into an enum rather than a string

t-young31:
### Checklist

### DIFF
--- a/api.web.yaml
+++ b/api.web.yaml
@@ -244,8 +244,7 @@ components:
         - kind
       properties:
         kind:
-          type: string
-          enum: [training_kind_nhsd]
+          $ref: "#/components/schemas/TrainingRecordKind"
         certificate_content_pdf_base64:
           type: string
           description: Base64 encoded PDF file data of the certificate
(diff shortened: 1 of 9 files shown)


api.web.yaml
```diff
@@ -321,3 +320,8 @@ components:
       type: array
       items:
         $ref: "#/components/schemas/Person"
+
+    TrainingRecordKind:
+      type: string
+      enum:
+        - "nhsd"
```
t-young31, line 327: suggestions welcome on whether this should be ''nhsd' or 'training_kind_nhsd'

jhughes982: - Although it is more verbose, I'd vote for `training_kind_nhsd` because implementations like `openapi.training_kind_nhsd` gives more context than `openapi.Nhsd`

jhughes982 (approved):
Looks good to me. Happy for the enum naming suggestion to go either way 👍

PR #145, opened 2025-07-01
approved researcher import

### LINKED ISSUE #141: Import approved researchers
## Description

**As a** infogov
**I want to** have all the approved researcher records available in one place for the TRE
**So that** I don't have to switch between systems

For migration we need the ability to import approved researcher records

## Acceptance criteria

- [ ] Records of user training and agreement acceptance can be imported
- [ ] e2e test added (inc. admin only ok)

t-young31:
## Resolves #141

- Adds a button to upload a csv file of approved researchers
- I went with runtime import as it's not _that_ much more work and makes multiple imports much easier

### Screenshots

Screenshot 2025-07-01 at 17 01 27

### Checklist

- [x] e2e test added

### DIFF
--- a/web/src/components/people/ApprovedResearcherImport.module.css
+++ b/web/src/components/people/ApprovedResearcherImport.module.css
@@ -5,6 +5,7 @@
 
   :link {
     color: blue;
+    text-decoration: underline;
   }
 }
 


web/src/components/people/ApprovedResearcherImport.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  async function handleFileUpload(e: React.ChangeEvent<HTMLInputElement>) {
+    const files = e.target.files;
+    if (!files) return;
+
+    const response = await postPeopleApprovedResearchersImportCsv({
+      body: files[0],
+    });
+    if (response.error) {
+      console.log(response.error);
```
acholyn, line 18:
```suggestion
      console.error(response.error);
```

web/src/components/people/ApprovedResearcherImport.tsx
```diff
@@ -0,0 +1,43 @@
...
+      setButtonText("Failed");
+    } else {
+      setButtonText("Imported ✔");
+    }
+    setButtonDisabled(true);
+  }
+
+  function handleSumbit(e: React.MouseEvent<HTMLButtonElement>) {
```
acholyn, line 26: I think this could actually be a `FormEvent` like what we use for [setting chosen name](https://github.com/ucl-arc-tre/portal/blob/4a333bc8df843543069b90a54b302c814d5786fb/web/src/components/profile/ProfileChosenName.tsx#L32C56-L32C56)

web/src/components/people/ApprovedResearcherImport.tsx
t-young31: the errors aren't user actionable so I'm inclined to not. if this was going to be used a lot then we'd send helpful error messages back from the API..

web/src/components/people/ApprovedResearcherImport.tsx
```diff
@@ -0,0 +1,62 @@
...
+          {buttonText}
+        </Button>
+      </form>
+      <span className={styles.helptext}>
+        Upload a .csv file (
+        <a href="https://github.com/ucl-arc-tre/portal/blob/cb4aaae44f270a943332b204adeafd73c23399a0/api.web.yaml#L156-L159">
+          format
+        </a>
+        ) containing approved researchers.
+      </span>
```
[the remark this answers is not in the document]
t-young31: fun. any ideas on doing it the correct (tooltip) way? (or any others?!) or could just add big ugly text 😆

web/src/components/people/ApprovedResearcherImport.tsx
jhughes982: Usually I would also be inclined to make this a bit more user friendly. For instance, the button becomes disabled on error and requires a manual refresh. But I guess this upload feature will become obsolete as soon as Sharepoint is replaced, so perhaps it's not worth adding additional features?

web/src/components/people/ApprovedResearcherImport.tsx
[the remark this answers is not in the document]
jhughes982: Perhaps this is one to run by Jason in our meeting later?

web/src/components/people/ApprovedResearcherImport.tsx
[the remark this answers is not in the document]
t-young31: for me that just inserts the child and does nothing on hover 😕

PR #159, opened 2025-07-07
Api endpoints refactor

### LINKED ISSUE #153: Update structure of Go handlers
- As I was reading through the Go handlers, I found some of them to be hard to understand at a quick glance because the function names don't match up to what the functions actually do. For instance:

```
// PostPeopleId updates a person's training record based on the provided user ID and training kind.
func (h *Handler) PostPeopleId(ctx *gin.Context, userId string) {
code...
}
```

- The above function might be more clear with a name like: `PutTrainingRecord`, or `PostUpdateTrainingRecord`, or `PostTrainingRecordById` (or another variation like that).
- Furthermore, the above handler function is currently in `people.go`. But since it focuses on training records I wonder if it would be better to have this in a dedicated `training.go` file?
- Then we could have handlers like so:
- `agreements.go` - focuses on methods related to agreements
- `auth.go` - focuses on auth
- `users.go` - would replace the current `people.go` and would focuses specifically on fetching and updating users
- `training.go` - focuses on posting, and updating training (even for specific users)

Thoughts welcome as to whether this is something people would find useful.

jhughes982:
## Resolves #153 

- Opening as draft for now while I fix tests
- This PR attempts to improve the structure and readability of Go handlers and services, api endpoints and schemas, and any associated typescript implementations.
- This is significantly driven by my own personal preferences. I think the new naming structure improves clarity but we can further discuss any areas that you disagree with.
- I've organised things such that we now have endpoints split into distinct areas (e.g. users, training, agreements, auth). So e.g., all training-related methods are structured under the `training` endpoint as opposed to the previous `people` endpoints. People has now been renamed to users.

### Checklist


api.web.yaml
```diff
@@ -23,166 +24,202 @@ paths:
         default:
           description: Unexpected error
 
-  /profile:
+  # User Management (Admin)
+  /users:
     get:
-      description: Users profile
+      operationId: getAllUsers
```
t-young31, line 30: I don't like this overriding. you need to remember both the path and what it's called

api.web.yaml
```diff
@@ -23,166 +24,202 @@ paths:
...
+                $ref: "#/components/schemas/Agreement"
+        "406":
+          description: Not acceptable
+        "500":
+          description: Internal server error
+        default:
+          description: Unexpected error
+
+  # Data Import (Admin)
+  /admin/approved-researchers/import/csv:
```
t-young31, line 220: I've not seen an API where the role you need is in the path, so I'm not a fan

api.web.yaml
```diff
@@ -224,7 +264,44 @@ components:
...
+        username:
+          type: string
+        id:
+          type: string
+
+    UserProfile:
+      type: object
+      description: Complete profile information for a user including identity, roles, agreements, and training status
+      required:
+        - user
```
t-young31, line 282: why does the user profile need the userId?

t-young31 (commented):
I like `/people` -> `/users`, but the rest not so much – sorry @jhughes982! (appreciate that this is pretty subjective though)

PR #160, opened 2025-07-08
possible refactor option

### LINKED ISSUE #153: Update structure of Go handlers

t-young31:
## Resolves #153 

Alternative to #159, which for me is clearer. Thoughts welcomed

### Checklist


t-young31:
> [...] we should make sure we are all clear on how this all works together. Some brief documentation could help with that.

strong agree!

jhughes982:
> > [...] we should make sure we are all clear on how this all works together. Some brief documentation could help with that.
> 
> strong agree!

Shall we organise a session where the three of us go through it together?

api.web.yaml
```diff
@@ -134,40 +141,42 @@ paths:
         default:
           description: Unexpected error
 
-  /people:
+  /users:
```
acholyn, line 144: also need to update nav with the url, do we want to call the page Users instead of People as well? Or do we think for researchers keeping People on the UI makes more sense?

acholyn: eh, let's leave it as people and see if it needs to be updated after user testing then

PR #168, opened 2025-07-10
add tre api

### LINKED ISSUE #161: Approved researcher API endpoint for TRE
**As a** TRE dev
**I want to** have an API that I can query for the approved researcher status & training expiry of a user
**So that** I can migrate from the sharepoint integration

## Acceptance criteria

- [ ] New API added
- [ ] Basic auth
- [ ] HLD updated

## Dependencies

 - [x] #151

t-young31:
## Resolves #161

- Refactors the API to allow top-level `/<service>` handlers
- Adds a `/tre` path with basic auth
- Drops the `web-` prefix from the API now the package handles both
- **Doesn't** use the entra controller to get the users yet, because I think it may not be required 🤦🏼
- Adds `NHSDTrainingExpiresAt` to the users service to get just the required value
  
### Checklist


jhughes982:
> @jhughes982 / @acholyn what're your thoughts on end-to-end testing this? pytest? go? cypress?

If there is a frontend component then cypress still seems ok. Otherwise writing Go tests should do the trick if it's purely backend?

jhughes982:
For the failing test I can see: `Error: API not available after multiple attempts. Your API Docker container may not be ready yet. Try again in 30 seconds`

This setting can be tweaked in `web/cypress/support/commands.ts`. The default is set at 5 attempts with a 2 second delay, but it can be anything you like. It seems like the api is taking a while to spin up, so increasing the attempts sounds reasonable. Otherwise, maybe you could expose a public `ping` route and have the other docker containers depend on a healthcheck of `ping`?

t-young31:
> `API not available after multiple attempts`

I think this is because it's missing some config, rather than being slow 😉

e2e/tre/main_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	baseUrl        = "http://api:8080/tre/api/v0"
+	username       = "username" // Must match e2e config values
+	password       = "password" // pragma: allowlist secret
+	requestTimeout = 1 * time.Second
+)
+
```
t-young31, line 18: should test unauthed too

jhughes982 (commented):
Looks good, although I think it needs some more documentation as the distinction between the two api's may not be so clear to new developers. Maybe just a short description about what the two focus on.

PR #172, opened 2025-07-16
Studies flow

jhughes982:
## Works towards #5 

- This is a continuation of the previous study and asset creation PR's. 
- The features are still unfinished, but I wanted to submit this PR as it completes one stage of the process (i.e. fetching and submitting studies). I think it will be easier to get feedback at this stage rather than carrying on with the assets flow and making the PR too big.
- So this PR just focusses on getting the flow set up for fetching and posting studies, including both the frontend and backend implementations.

## Frontend
- We now have a logical flow where users are shown their list of studies (future work will allow users to select and manage a study).
- Users can now create a new study and post it to the database.
- In order to get the studies form submission to work, I needed to modify certain elements of the studies form. I'm still not entirely sure if I have covered everything, so I would appreciate a deeper look from @acholyn on that one.
- A lot of the `asset` related logic has been commented out temporarily while I get the studies set up first. It's a bit messy but will hopefully be cleaned up in the next PR.

## Backend
- Backend logic has been added to set up the study schema in the database.
- Endpoints and services have also been added to allow form submission and fetching from the database.
- For the schemas, endpoints, and services I am very happy to modify any of these further.
- Theres also some RBAC stuff that I think needs adding. I'll need to think more about that and maybe have further conversations to figure out whats required.
- I also extended the `setErrResponse` to make it easier to read error messages in the response object.

## Testing
- The form tests had already been added as part of the original. I can add more once I get an initial round of feedback on this.

### Checklist


api.web.yaml
```diff
@@ -542,3 +584,134 @@ components:
...
+        - data_protection
+        - data_protection_prefix
+        - data_protection_date
+        - data_protection_id
+        - data_protection_number
+        - third_party
+        - external_users
+        - consent
+        - non_consent
+        - ext_eea
```
jhughes982, line 622: - I have set these to `required` by default as I thought it would be better to have these explicitly set in the database (i.e. even if they are empty strings or `false` by default).

- But we can change this if desired so that optional fields do not get set in the database if they are not explicitly set by the user.

internal/handler/web/studies.go
```diff
@@ -7,9 +7,109 @@ import (
...
+			DataProtectionDate:   study.DataProtectionDate,
+			DataProtectionId:     study.DataProtectionId,
+			DataProtectionNumber: study.DataProtectionNumber,
+			ThirdParty:           study.ThirdParty,
+			ExternalUsers:        study.ExternalUsers,
+			Consent:              study.Consent,
+			NonConsent:           study.NonConsent,
+			ExtEea:               study.ExtEea,
+			CreatedAt:            study.CreatedAt.Format("2006-01-02T15:04:05Z07:00"),
+			UpdatedAt:            study.UpdatedAt.Format("2006-01-02T15:04:05Z07:00"),
```
jhughes982, line 57: Let me know if there is a cleaner way to do this.

web/src/components/assets/AssetForm.tsx
```diff
@@ -423,7 +423,12 @@ export default function AssetForm(props: AssetFormProps) {
             {isSubmitting ? "Creating..." : "Create Asset"}
           </Button>
 
-          <Button type="button" variant="secondary" onClick={() => setSelectedStudy(null)} disabled={isSubmitting}>
+          <Button
+            type="button"
+            variant="secondary"
+            // onClick={() => setSelectedStudy(null)}
+            disabled={isSubmitting}
```
jhughes982, line 430: There are some comments in this PR while I set things up in stages.

web/src/pages/assets/index.tsx
```diff
@@ -29,18 +29,19 @@ export default function AssetsPage() {
...
+        //     id: "1",
+        //     title: "COVID-19 Research Study",
+        //     description: "A comprehensive study on COVID-19 impacts and treatments",
+        //   },
+        //   {
+        //     id: "2",
+        //     title: "Cancer Treatment Analysis",
+        //     description: "Analysis of various cancer treatment methodologies",
+        //   },
+        // ]);
```
jhughes982, line 44: As said above, assets is a bit messy, just until I get this first stage out of the way.

jhughes982:
@t-young31 any idea what the lint failure is about?

t-young31 (commented):
looking good – some initial comments ^

api.web.yaml
```diff
@@ -542,3 +584,134 @@ components:
...
+        owner_user_id:
+          type: string
+          description: ID of the user who owns the study
+        admin:
+          type: string
+          description: Email of the study administrator
+        controller:
+          type: string
+          enum: ["UCL", "Other"]
+          description: Data controller organization
```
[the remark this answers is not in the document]
jhughes982: Again, @acholyn 's ideas would be good here? Happy to go with whatever makes most sense.

api.web.yaml
```diff
@@ -542,3 +584,134 @@ components:
...
+        controller_other:
+          type: string
+          description: Other data controller if controller is "Other"
+        ucl_sponsorship:
```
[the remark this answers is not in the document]
jhughes982: I agree. I can start to change these unless @acholyn wanted to carry on with the form attributes?

internal/handler/web/responses.go
```diff
@@ -20,6 +20,9 @@ func setErrResponse(ctx *gin.Context, err error, status int, message string) {
 	user := middleware.GetUser(ctx)
 	if err != nil {
 		log.Err(err).Any("username", user.Username).Msg(message)
-		ctx.Status(status)
+		ctx.JSON(status, gin.H{
+			"error":   message,
+			"details": err.Error(),
+		})
```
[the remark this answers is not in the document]
jhughes982: Happy to revert this back if it's a concern?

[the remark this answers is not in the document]
t-young31: safety first 👍🏼

api.web.yaml
```diff
@@ -542,3 +584,134 @@ components:
         updated_at:
           type: string
           description: Time in RFC3339 format when the asset was last updated
+
+    Study:
+      type: object
+      description: A research study
+      required:
+        - id
```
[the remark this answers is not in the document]
jhughes982: Some more thoughts: since the `id`, `created_at`, `updated_at`, and `OwnerUserID` are set on the backend, should we have a separate response model for the study `POST` action that omits these fields? Something like: `$ref: "#/components/schemas/StudyCreateRequest"` which would be separate from the `study` schema itself?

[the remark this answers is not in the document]
t-young31: > Should we have them all required and simply default to "" and false unless they are explicitly set by the user?

I think optional fields should be null. e.g. to me `false` implies the user has chosen `false`

> should we have a separate response model for the study

sounds good. maybe the shared bits in a `StudyAttributes` object? (not fussed about splitting on the frontend)

internal/service/studies/main.go
```diff
@@ -17,6 +17,29 @@ func New() *Service {
 	}
 }
 
+func (s *Service) CreateStudy(userID uuid.UUID, studyData types.Study) (*types.Study, error) {
+	studyData.OwnerUserID = userID
+
```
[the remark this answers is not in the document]
jhughes982: https://github.com/ucl-arc-tre/portal/pull/172/commits/80f94b64d905c249394b9b24388dbebfbd840bfc

Is this what you had in mind? Happy to be advised on the best way of handling the errors in Go.

api.web.yaml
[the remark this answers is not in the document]
jhughes982: https://github.com/ucl-arc-tre/portal/pull/172/commits/1b2ad7ba0c032ebfb28366b6afda0a376fb7eda1

I added the suggestions in the above commit. So now only the `title` and `controller` fields are required and the rest of the attributes default to `null` in the database. Let me know if this approach is better. As usual, guidance on the Go code is welcome.

internal/service/studies/main.go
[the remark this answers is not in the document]
t-young31: #174

web/src/components/studies/CreateStudyForm.tsx
```diff
@@ -370,29 +393,37 @@ export default function CreateStudyForm(CreateStudyProps: CreateStudyProps) {
...
+                      inputClassName={controllerValue === "UCL" ? styles.readonly : ""}
+                    />
+                  )}
+                />
+
+                <Controller
+                  name="dataProtectionDate"
+                  control={control}
+                  render={({ field }) => (
+                    <input
```
acholyn, line 415: is there a reason this isn't also a UI kit component?

web/src/components/studies/CreateStudyForm.tsx
```diff
@@ -401,8 +432,18 @@ export default function CreateStudyForm(CreateStudyProps: CreateStudyProps) {
...
-                  })}
-                  placeholder="eg 123"
+                  }}
+                  render={({ field }) => (
+                    <input
+                      {...field}
+                      type="number"
+                      id="dataProtectionId"
+                      placeholder="eg 123"
+                      style={{ padding: "0.5rem", fontSize: "1rem", border: "1px solid #ccc", borderRadius: "4px" }}
```
acholyn, line 442: if we're rounding the outer edges, we probably also want to do this for the registry ID too

t-young31: I think UCL style is no rounding – so we should be consistent and have none anywhere(?)

jhughes982: https://github.com/ucl-arc-tre/portal/pull/172/commits/9d33c7fbfbaca2a32c7bb6c4e29dd440d9fa15b6

I removed the border radius and inline styles 👍

web/src/components/studies/CreateStudyForm.tsx
jhughes982: For some reason the data was not getting saved when using the uikit components. Happy for you to dig into it to see if you can get it working though.

api.web.yaml
```diff
@@ -542,3 +584,134 @@ components:
...
+          description: Title of the study
+        description:
+          type: string
+          description: Description of the study
+        owner_user_id:
+          type: string
+          description: ID of the user who owns the study
+        admin:
+          type: string
+          description: Email of the study administrator
```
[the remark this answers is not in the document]
acholyn: I didn't realise they were allowed to nominate a number of them upon creation, if they are then I agree with Tom, I'd specify it's a comma separated list (or we have multiple input boxes and combine them into one field like with the `dataProtectionNumber` field
we'd then need to validate they're all UCL staff (maybe when user clicks next/stops inputting? or else just on submit and show it as an error)

api.web.yaml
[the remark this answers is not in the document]
acholyn: I don't disagree, but I'm a little hazy on the distinction of if you can have a data controller at the study level or if it really is only on the asset level, @finleybacon , any thoughts?

api.web.yaml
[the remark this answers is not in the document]
acholyn: nah, happy for you to update them; I only copied over what was in the original for simplicity

jhughes982:
https://github.com/ucl-arc-tre/portal/pull/172/commits/a8d067c3046ed5207a794943d952ed44cfda7d3b

I added some initial RBAC policies for approved researchers having access to studies and assets. @t-young31 could you give this a quick review and let me know if I missed anything.

Should the base role have access to `GET` studies and assets?

acholyn:
> Should the base role have access to `GET` studies and assets?

Think they should be able to get the studies they're part of and see the assets within. We might need some security tweaks on assets later on if some things should only be viewable by certain clearance or something, but I don't think we need to worry about that now

t-young31:
> Should the base role have access to GET studies and assets?

yes in the future, now – in my opinion – no (principal of least privilege)

jhughes982:
Question: should the study description be `required`?

acholyn:
> Question: should the study description be `required`?

Not sure if there's much benefit to making it required; maybe some titles are very straightforward, and they have to be unique so a description won't really be used to distinguish 2 studies with the same name. I think encouraged but not required  in the initial form

finleybacon:
> > We will be seeking/have sought approval from the Confidentiality Advisory Group for this research
> 
> We could set this as `has_cag_approval` - but the question allows some leeway here in that the user can set this to `true` if they just 'intend' to seek the approval but do not actually have it yet. Should we reword this to ask whether they have the approval or not more explicitly?
> 
> I think there are other variables like this too. So we should be consistent in the wording.

@jhughes982 Agree. "Will you be seeking/have you sought" is quite common in the current form. I don't know why it's like that, although you can image the leeway is helpful in some use cases. I'm wary of making too many things required or explicit on creation because I think it's helpful to allow people to create a bare-bones Study initially, and then come back and add in more data as and when it comes through - some of these approvals take weeks/months to trickle through various committees.

So if we keep the leeway (or remove the leeway from the questions but make them not required), we'll need some validation/review step later on which ensures the Study has the minimum set of data before a Project can be created.

jhughes982:
> @jhughes982 Agree. "Will you be seeking/have you sought" is quite common in the current form. I don't know why it's like that, although you can image the leeway is helpful in some use cases. I'm wary of making too many things required or explicit on creation because I think it's helpful to allow people to create a bare-bones Study initially, and then come back and add in more data as and when it comes through - some of these approvals take weeks/months to trickle through various committees.
> 
> So if we keep the leeway (or remove the leeway from the questions but make them not required), we'll need some validation/review step later on which ensures the Study has the minimum set of data before any Projects can be created.

Thanks Finley! It sounds like the easiest approach for now is to keep the original wording and I'll just come up with an appropriate variable name in the database.

jhughes982:
I think I've addressed this first round of comments now. I propose we merge this in if everyone is happy and I can move on to the assets.

One thing that might be worthwhile is for @finleybacon to review the database columns for the `study` and make sure we are capturing everything thats needed (we can also remove anything thats not needed at this stage). This is what the current columns are:

```
type Study struct {
	ModelAuditable
	OwnerUserID                      uuid.UUID `gorm:"not null;index"`
	Title                            string    `gorm:"not null"`
	Description                      *string   `gorm:"type:text"`
	Admin                            *string   `gorm:"type:varchar(255)"`
	Controller                       string    `gorm:"not null"`
	ControllerOther                  *string   `gorm:"type:varchar(255)"`
	InvolvesUclSponsorship           *bool     `gorm:""`
	InvolvesCag                      *bool     `gorm:""`
	CagReference                     *string   `gorm:"type:varchar(255)"`
	InvolvesEthicsApproval           *bool     `gorm:""`
	InvolvesHraApproval              *bool     `gorm:""`
	IrasId                           *string   `gorm:"type:varchar(255)"`
	IsNhsAssociated                  *bool     `gorm:""`
	InvolvesNhsEngland               *bool     `gorm:""`
	NhsEnglandReference              *string   `gorm:"type:varchar(255)"`
	InvolvesMnca                     *bool     `gorm:""`
	RequiresDspt                     *bool     `gorm:""`
	RequiresDbs                      *bool     `gorm:""`
	IsDataProtectionOfficeRegistered *bool     `gorm:""`
	DataProtectionPrefix             *string   `gorm:"type:varchar(255)"`
	DataProtectionDate               *string   `gorm:"type:varchar(255)"`
	DataProtectionId                 *int      `gorm:""`
	DataProtectionNumber             *string   `gorm:"type:varchar(255)"`
	InvolvesThirdParty               *bool     `gorm:""`
	InvolvesExternalUsers            *bool     `gorm:""`
	InvolvesParticipantConsent       *bool     `gorm:""`
	InvolvesIndirectDataCollection   *bool     `gorm:""`
	InvolvesDataProcessingOutsideEea *bool     `gorm:""`

	// Relationships
	Owner  User    `gorm:"foreignKey:OwnerUserID"`
	Assets []Asset `gorm:"foreignKey:StudyID"`
}
```

internal/rbac/init.go
```diff
@@ -36,6 +37,18 @@ func addBasePolicies(enforcer *casbin.Enforcer) {
 	}
 }
 
+func addApprovedResearcherPolicies(enforcer *casbin.Enforcer) {
+	policies := []Policy{
+		{RoleName: ApprovedResearcher, Resource: "/studies", Action: ReadAction},
+		{RoleName: ApprovedResearcher, Resource: "/studies", Action: WriteAction},
+		{RoleName: ApprovedResearcher, Resource: "/studies/*/assets", Action: ReadAction},
+		{RoleName: ApprovedResearcher, Resource: "/studies/*/assets", Action: WriteAction},
```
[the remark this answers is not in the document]
jhughes982: https://github.com/ucl-arc-tre/portal/pull/172/commits/19a8a69c7f180cdec67cac8b022a597da291b342

Implementation commented out for now

api.web.yaml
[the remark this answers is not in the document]
jhughes982: Also, George wanted to avoid columns that contain multiple values and instead have all the values contained in a separate DB table. Is that the approach to use here?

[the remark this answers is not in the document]
t-young31: I think we should ask for usernames (would be nice to add the lookup functionality for the email but is a nice-to-have, in my opinion)

internal/service/studies/main.go
```diff
@@ -17,6 +21,83 @@ func New() *Service {
 	}
 }
 
+func (s *Service) CreateStudy(userID uuid.UUID, studyData openapi.StudyCreateRequest) (*types.Study, error) {
+	if strings.TrimSpace(studyData.Title) == "" {
+		return nil, errors.New("study title is required")
```
t-young31, line 26: I think we should come up with some regex for a valid study name and enforce it on the frontend and backend. e.g `^\w[\w\s\-]{2,48}\w$`

internal/service/studies/main.go
```diff
@@ -17,6 +21,83 @@ func New() *Service {
...
+	}
+
+	if studyData.DataControllerOrganisation == nil || strings.TrimSpace(string(*studyData.DataControllerOrganisation)) == "" {
```
t-young31, line 29: if this is required then it can be a required field and we just check that it's not `""` here(?)

internal/service/studies/main.go
t-young31, line 26: can we here and below use the new https://github.com/ucl-arc-tre/portal/blob/325369a5aafc6e3a1607623b490dc9817833a635/internal/types/errors.go#L13 error so the handler returns the the correct response (and equiv for server error)

internal/service/studies/main.go
```diff
@@ -17,6 +21,83 @@ func New() *Service {
...
+	dbStudy.InvolvesParticipantConsent = studyData.InvolvesParticipantConsent
+	dbStudy.InvolvesIndirectDataCollection = studyData.InvolvesIndirectDataCollection
+	dbStudy.InvolvesDataProcessingOutsideEea = studyData.InvolvesDataProcessingOutsideEea
+
+	if err := s.db.Create(&dbStudy).Error; err != nil {
+		return nil, err
+	}
+
+	return &dbStudy, nil
+}
```
t-young31, line 86: this function is a little long for my taste – potentially consider refactoring the validation into a `validateCreate(study openapi.StudyCreateRequest) error`

internal/types/studies.go
```diff
@@ -6,9 +6,34 @@ import (
 
 type Study struct {
 	ModelAuditable
-	OwnerUserID uuid.UUID `gorm:"not null;index"`
-	Title       string    `gorm:"not null"`
-	Description string    `gorm:"type:text"`
+	OwnerUserID                      uuid.UUID `gorm:"not null;index"`
+	Title                            string    `gorm:"not null"`
```
t-young31, line 10: nit: `Name` might be better if that's what people are used to

internal/types/studies.go
```diff
@@ -6,9 +6,34 @@ import (
...
+	Description                      *string   `gorm:"type:text"`
+	Admin                            *string   `gorm:"type:varchar(255)"`
+	Controller                       string    `gorm:"not null"`
+	ControllerOther                  *string   `gorm:"type:varchar(255)"`
```
t-young31, line 14: I'm not a big fan of this. querying will be more complicated than it needs to be(?). all of `ucl`, `Ucl` and `UCL` `University college london` should be stored as the same thing

internal/types/studies.go
```diff
@@ -6,9 +6,34 @@ import (
...
+	InvolvesUclSponsorship           *bool     `gorm:""`
+	InvolvesCag                      *bool     `gorm:""`
+	CagReference                     *string   `gorm:"type:varchar(255)"`
```
t-young31, line 17: these and some of the other fields I think warrant inline comments to say what they are

internal/types/studies.go
```diff
@@ -6,9 +6,34 @@ import (
...
+	InvolvesNhsEngland               *bool     `gorm:""`
+	NhsEnglandReference              *string   `gorm:"type:varchar(255)"`
+	InvolvesMnca                     *bool     `gorm:""`
+	RequiresDspt                     *bool     `gorm:""`
+	RequiresDbs                      *bool     `gorm:""`
+	IsDataProtectionOfficeRegistered *bool     `gorm:""`
+	DataProtectionPrefix             *string   `gorm:"type:varchar(255)"`
+	DataProtectionDate               *string   `gorm:"type:varchar(255)"`
+	DataProtectionId                 *int      `gorm:""`
+	DataProtectionNumber             *string   `gorm:"type:varchar(255)"`
```
t-young31, line 31: what was the conclusion on these? I'm tempted to say just the `DataProtectionNumber`. pattern matching in postgres is fine so querying some substring of it should be fine(?)

internal/service/studies/main.go
acholyn: field should already be required so it should throw an error if it doesn't have any content (have checked and it does do this), shouldn't the DataControllerOrganisation == "Other" not nil?

t-young31: doesn't look like it  – might have been missed in a refactor https://github.com/ucl-arc-tre/portal/blob/8075505a2369c8f5ee6c0619f4f1137838d3568b/api.web.yaml#L591-L593C11

api.web.yaml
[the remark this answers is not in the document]
acholyn: worth the effort though, surely? Think it might largely depend on the entra methods available

internal/service/studies/main.go
```diff
@@ -1,27 +1,111 @@
...
+	}
+	if err := s.db.Create(&user).Error; err != nil {
+		return nil, fmt.Errorf("error creating user: %w", err)
+	}
+
+	return &user, nil
+}
+
+// validate all admin usernames and create/find corresponding users
+func (s *Service) validateAndCreateAdmins(ctx context.Context, usernames []string) ([]types.User, error) {
```
t-young31, line 79: this I think would be more appropriate in the users service...(?)

internal/service/studies/main.go
```diff
@@ -1,27 +1,111 @@
...
+
+	// if *userData.EmployeeType != "staff" {
+	// 	return fmt.Errorf("username '%s' is not a valid staff member", username)
+	// }
+
+	return nil
+}
+
+// find a user in the database or create one if it doesn't exist
+func (s *Service) findOrCreateUser(username string) (*types.User, error) {
```
t-young31, line 56: this is duplicated in the users service

internal/handler/web/studies.go
```diff
@@ -26,12 +26,23 @@ func (h *Handler) GetStudies(ctx *gin.Context) {
...
+
+		// Extract admin usernames from StudyAdmins
+		var adminUsernames []string
+		for _, studyAdmin := range study.StudyAdmins {
+			adminUsernames = append(adminUsernames, string(studyAdmin.User.Username))
+		}
+		var adminUsernamesPtr *[]string
+		if len(adminUsernames) > 0 {
+			adminUsernamesPtr = &adminUsernames
+		}
```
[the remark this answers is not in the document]
jhughes982: As far as I am aware it is not a required field, it's optional.

jhughes982:
> can also remove IsNhsAssociated since this was just to show the other NHS related questions, unless @finleybacon this is helpful?

I'm wondering whether there is risk of loosing valuable data here if we remove the `IsNhsAssociated` checkbox from the database. Consider the following:

If the user clicks the box that targets `IsNhsAssociated` they do not necessarily have to click any of the conditional checkboxes that appear afterwards. So If we remove this field and the user ends up clicking the first checkbox (thereby triggering `IsNhsAssociated` to be `true`) but then they don't click any of the conditional checkboxes we may end up missing the information that their study is associated with the NHS.

@finleybacon @acholyn @t-young31 any thoughts?

t-young31:
happy to keep it

finleybacon:
> If the user clicks the box that targets `IsNhsAssociated` they do not necessarily have to click any of the conditional checkboxes that appear afterwards

Maybe keep the column but review this ^ at a later time, since I wonder what utility anyone would get from the claim "Study X is NHS associated" with no further info on how it is associated.

jhughes982:
> Maybe keep the column but review this ^ at a later time, since I wonder what utility anyone would get from the claim "Study X is NHS associated" with no further info on how it is associated.

One approach would be to add validation so that if the user clicks the `Study X is NHS associated` checkbox they have to follow up with at least one of the conditional boxes. 

I didn't want to make this PR any longer than it already is, but since I am here now I might as well address these final few points 😅 

What does everyone think?

jhughes982:
Ok, I have addressed all of the current comments that would block this from being merged. I'd encourage everyone to carefully check this and have a play around as the PR got very large and I may have missed some edge-cases.

There are a few comments that do still need to be addressed but they can be tackled in a separate PR (e.g. using the correct error types). 

Here are some additional observations I have made which I think are critical:
1. The error handling should be improved (in follow-up PR's) to provide better UX. For instance, if an error is found with the additional study admins the frontend will show a generic error but the specific details don't get returned from the backend.
2. The entra logic needs to be discussed further as I ran into a few issues. We discussed earlier that the additional study admins should take a username, but I think the entra service is querying a column called `User principal name` which as far as I can see on our `ARC.TRE.TEST` account contains emails. 
3. Following from point 2, the users currently in `ARC.TRE.TEST` do not have an `employee type` populated (at least not for the ones I checked) so I can't fully test out the logic that validates if they are a staff member or not. For now this logic has been commented out.

If any of the above points are actually critical to address before we merge this I can work on it further. Otherwise I'd like to follow up in smaller PR's.

web/src/components/studies/CreateStudyForm.tsx
```diff
@@ -353,15 +380,16 @@ export default function CreateStudyForm(CreateStudyProps: CreateStudyProps) {
...
 
           <Label htmlFor="dataProtection" className={styles["checkbox-label"]}>
-            <input type="checkbox" id="dataProtection" {...register("dataProtection")} />
+            <input type="checkbox" id="dataProtection" {...register("isDataProtectionOfficeRegistered")} />
             The research is already registered with the UCL Data Protection Office
           </Label>
+
           {showDataProtectionNumber && (
             <Label htmlFor="dataProtectionNumber">
               Data Protection Registration Number:
```
acholyn, line 395:
```suggestion
              Data Protection Registration Number*:
```

PR #176, opened 2025-07-18
update approved researcher status on training update

### LINKED ISSUE #175: Manual Training Approval doesn't update role
Presently the only way to get approved researcher status is through uploading a valid training certificate; manually approving the training doesn't update the roles provided to the user. The user has valid training but there isn't a mechanism that checks for valid training + agreement and sets the role.

t-young31:
### Checklist


jhughes982 (approved):
Looks good.

PR #181, opened 2025-07-22
agreement encouragement

### LINKED ISSUE #81: Firm encouragement of agreement reading
## Promote user understanding and engagement with agreement through UI
**As a** developer
**I want to** ensure the user reads and understands the approved researcher agreement
**So that** we can trust our users and avoid avoidable problems

## Acceptance criteria

- [ ] Agreement should come after training upload
- [ ] Agreement text should be restructured so the bullet points can come under subheadings/in chunks (this might include adding in section IDs for better fetch & display)
- [ ] Display the agreement in an interactive/split up manner, where they agree and continue for each section
  - [ ] Display a message like "Please read each statement carefully before you agree and move on"
  - [ ] Show the initial blurb (non-bulleted) of the agreement in a card, give eg 1 min with the timer in the corner of the card until it is done and then the "agree and continue" button becomes live
  - [ ] Repeat for the sections/chunks of bullet points
    - [ ] Last card should have "agree and close"
- [ ] Show confirmation message at end and mention users can review the agreement from their profile

## Dependencies

- updating agreement structure

t-young31:
## Resolves #81 

- Adds a wait timer before the approved researcher agreement can be agreed to
- Tweaks the styling on the approved researcher process to be more consistent (factoring out @jhughes982's box)
- Adds a build-time config for the frontend to allow a short required wait time in dev (because I'm impatient!). Thoughts on a better way to do this appreciated

### Screenshots

Screenshot 2025-07-22 at 14 14 03
Screenshot 2025-07-22 at 14 14 10

### Checklist


web/src/config.ts
```diff
@@ -0,0 +1 @@
+export const approvedAgreementRequiredReadingSeconds = 90;
```
t-young31, line 1: thoughts on naming and what this number should be are very welcome

web/src/components/profile/approved-researcher/AgreementForm.tsx
```diff
@@ -33,22 +45,15 @@ export default function ApprovedResearcherForm(props: ApprovedResearcherFormProp
...
-              setAgreed(!agreed);
-            }}
-            checked={agreed}
-            required
-            aria-label="I agree to the approved researcher agreement"
-          />
-          I agree
-          <Button size="large" type="submit" disabled={!agreed} cy="approved-researcher-agreement-agree">
-            Submit
+          <Button size="large" type="submit" disabled={!canAgree} cy="approved-researcher-agreement-agree">
```
acholyn, line 55: might want to add the aria-label the checkbox had

web/src/components/profile/approved-researcher/AgreementForm.module.css
```diff
@@ -1,7 +1,6 @@
 .wrapper {
   max-width: 400px;
-  margin: 2rem auto;
-  padding: 1rem;
+  margin: 1rem auto;
 }
 
 .checkbox {
```
acholyn, line 6: we can probably get rid of the checkbox styling

web/src/config.ts
acholyn: >That would be really nice – any ideas how within the constraints of uikit?

maybe something like (assuming we add `className={styles["agreement-button"]}`
```css
.agreement-button {
transition: background-color 90s ease;
background-color: #0d68cf;

&:disabled {
 background-color: #808080; or var(--grey)
}
}
```

acholyn: won't allow us to do anything when it's actually disabled, so we could set the disabled state for 60-80s and the fade in will happen after that, which can take 30-40s and will give the illusion of being disabled for 90-100s

```css
.agreement-button {
  width: 100%;
  transition: background-color 30s ease;
  background-color: #0d68cf;
}
```

t-young31: @jhughes982 what do you think?

deploy/dev/docker-compose.yaml
```diff
@@ -32,6 +32,7 @@ services:
       - type: bind
         source: ../..
         target: /repo
+      - ./config.ts:/repo/web/src/config.ts:ro
```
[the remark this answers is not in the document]
t-young31: I want https://github.com/ucl-arc-tre/portal/blob/d80cfea9861e9eb9fef9a182d565c7beb75c778e/deploy/dev/config.ts#L1 to be build-time configurable (so don't need to wait so long in dev). better ideas appreciated!

web/src/components/profile/ProfileStepProgress.tsx
```diff
@@ -15,7 +16,7 @@ export default function ProfileStepProgress(props: ProfileStepProgressProps) {
   const { steps, profileIsComplete } = props;
 
   return (
-    <div className={styles.container}>
+    <Box>
```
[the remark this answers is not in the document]
t-young31: also https://github.com/ucl-arc-tre/portal/blob/d80cfea9861e9eb9fef9a182d565c7beb75c778e/web/src/components/profile/approved-researcher/ApprovedResearcherAgreement.tsx#L66 and https://github.com/ucl-arc-tre/portal/blob/d80cfea9861e9eb9fef9a182d565c7beb75c778e/web/src/components/profile/approved-researcher/TrainingCertificate.tsx#L153

web/src/components/profile/approved-researcher/AgreementForm.tsx
```diff
@@ -16,15 +15,28 @@ type ApprovedResearcherFormProps = {
...
+    const timer = setInterval(() => {
+      setSecondsRemaining((secondsRemaining) => {
+        if (secondsRemaining === 0) {
+          clearInterval(timer);
+          return 0;
+        }
+        return secondsRemaining - 1;
+      });
+    }, 1000);
+  }, []);
```
jhughes982, line 32: This should have a cleanup function added to avoid memory leaks, something like:

```
  useEffect(() => {
    if (secondsRemaining === 0) return;

    const timer = setTimeout(() => {
      setSecondsRemaining((prev) => prev - 1);
    }, 1000);

    return () => clearTimeout(timer);
  }, [secondsRemaining]);
```

deploy/dev/docker-compose.yaml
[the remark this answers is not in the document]
jhughes982: The env variable will get baked into the code at build time which would then be available to the client when they access the app, so it should work with a static export. That means you'd need to be ok with the env variable being public. You'd use it like this:


```
env.production

// the NEXT_PUBLIC prefix should make it work with a static export
NEXT_PUBLIC_AGREEMENT_TIMER=60
```


```
AgreementForm.tsx

const agreementTimer = process.env.NEXT_PUBLIC_AGREEMENT_TIMER;
```

The advantage here is that I think you'd only need to declare it once in the nextJS codebase. Then when Docker builds the app it should get populated during `npm run build`.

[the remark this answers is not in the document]
t-young31: are you sure? tests are 💀 due to
Screenshot 2025-07-24 at 14 53 34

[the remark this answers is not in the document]
jhughes982: It should work, I just ran `npm run dev` locally on your branch and the variable was picked up successfully. It also worked after running `npm run build` and then serving it locally (without docker).

 Perhaps there is something else we need to do to get it to work with Docker?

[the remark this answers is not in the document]
t-young31: I'm confused – does the auth work with that? also the value isn't in the html?!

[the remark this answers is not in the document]
jhughes982: The auth won't work as it's just exposing the frontend (for testing). I'm also not sure why it's not searchable in the frontend. Will have to have a think.

You could always just stick with your `config.ts` if this drags on.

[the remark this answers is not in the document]
jhughes982: I didn't. I logged the env variable on the homepage.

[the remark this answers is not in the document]
jhughes982: Good that it's working now. You could add a `.env.example` to the repo, and then I can also reuse that for my [recent suggestion](https://github.com/ucl-arc-tre/portal/pull/185#discussion_r2228625647).

web/.gitignore
```diff
@@ -15,12 +15,6 @@
 /dist
 /dist-ssr
 
-
-# Environment
-.env*
-.env.local
-.env.*.local
-
```
jhughes982, line 23: You would usually ignore `.env.production`. Are we not ignoring it here because we don't intend for it to hold any sensitive information? A wonder if a brief explanation of this in the README would be good so that new devs have that context?

web/package.json
```diff
@@ -3,8 +3,8 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "next dev --turbopack",
-    "build": "next build",
+    "dev": "NODE_ENV=development next dev --turbopack",
+    "build": "NODE_ENV=production next build",
```
jhughes982, line 7: I thought NextJS was able to pick up the `NODE_ENV` automatically based on whether you are running `npm run dev` vs. `npm run build`, but maybe I am mistaken?

web/.gitignore
t-young31: > Are we not ignoring it here because we don't intend for it to hold any sensitive information?

it cannot hold any sensitive information because it's all public static frontend..

> A wonder if a brief explanation of this in the README would be good so that new devs have that context?

quite possibly. suggestions welcomed

jhughes982: Maybe have an environment variables section in the `web/README.md` and say something like:

```
## Environment variables

This project checks in to git both the `.env.development` and `.env.production` files. Make sure not to include any sensitive information in either of the environment variable files. This is because the frontend is statically exported and so all environment variables will be publicly available. Feel free to use environment variables for convenient config setup.
```

acholyn (approved):

jhughes982 (approved):
Looks good

PR #198, opened 2025-07-30
Initial external entra invitations

acholyn:
### Checklist
- adds a route to invite externals
- adds a button + form to the admin view of People to invite externals
- Go logic to use entra to send out an invite (will only work for those who have not previously accepted an invite)
- added redirect url to dev config and entra credentials

image
image
image

---

I'm thinking to follow this up with the checks for if the user is in our db/tenant or assign sponsors (depending on if the email account is ready)


internal/config/main.go
```diff
@@ -67,6 +67,7 @@ func EntraCredentials() EntraCredentialBundle {
 		TenantID:     k.String("entra.tenant_id"),
 		ClientID:     k.String("entra.client_id"),
 		ClientSecret: k.String("entra.client_secret"),
+		RedirectURL:  k.String("entra.redirect_url"),
```
t-young31, line 70: this isn't a credential so shouldn't be here(?)

deploy/dev/config.sample.yaml
```diff
@@ -8,6 +8,7 @@ entra:
   tenant_id: <CHANGE_ME>
   client_id: <CHANGE_ME>
   client_secret: <CHANGE_ME>
+  redirect_url: <CHANGE_ME>
```
t-young31, line 11: mind adding an example of this? (I'm not sure what this should be!)

internal/controller/entra/main.go
```diff
@@ -97,3 +98,33 @@ func (c *Controller) IsStaffMember(ctx context.Context, username types.Username)
...
+	requestBody := graphmodels.NewInvitation()
+	invitedUserEmailAddress := email
+	inviteRedirectUrl := credentials.RedirectURL
+	sendInvitationMessage := true
+
+	requestBody.SetInvitedUserEmailAddress(&invitedUserEmailAddress)
+	requestBody.SetInviteRedirectUrl(&inviteRedirectUrl)
+	requestBody.SetSendInvitationMessage(&sendInvitationMessage)
+
+	message := "You have been invited to join the UCL ARC Portal by " + string(sponsor.Username)
```
t-young31, line 114: only the username of the sponsor is used, so I'd go for

```go
func (c *Controller) SendInvite(ctx context.Context, email string, sponsorUsername types.Username) error {
```

internal/controller/entra/main.go
```diff
@@ -97,3 +98,33 @@ func (c *Controller) IsStaffMember(ctx context.Context, username types.Username)
...
+	messageInfo := graphmodels.NewInvitedUserMessageInfo()
+
+	messageInfo.SetCustomizedMessageBody(&message)
+	requestBody.SetInvitedUserMessageInfo(messageInfo)
+
+	// todo: set sponsor (separate func)
```
t-young31, line 120: rm?

internal/controller/entra/main.go
```diff
@@ -97,3 +98,33 @@ func (c *Controller) IsStaffMember(ctx context.Context, username types.Username)
...
+
+	_, err := c.client.Invitations().Post(ctx, requestBody, nil)
+	if err != nil {
+		return fmt.Errorf("failed to send invite to '%s': %v", invitedUserEmailAddress, err)
```
t-young31, line 124: server error type

internal/controller/entra/main.go
```diff
@@ -97,3 +98,33 @@ func (c *Controller) IsStaffMember(ctx context.Context, username types.Username)
...
+	}
+
+	log.Printf("Invitation sent: %+v\n", invitedUserEmailAddress)
```
t-young31, line 127: zerolog please

internal/handler/web/users.go
```diff
@@ -89,3 +89,30 @@ func (h *Handler) PostUsersApprovedResearchersImportCsv(ctx *gin.Context) {
...
+	attributes, err := h.users.Attributes(user)
+	if err != nil {
+		setError(ctx, err, "Failed to get user attributes")
+		return
+	}
+
+	sponsor := types.Sponsor{
+		Username:   user.Username,
+		ChosenName: attributes.ChosenName,
+	}
```
t-young31, line 104: can rm if just providing the username of the sponsor

internal/service/users/invite.go
```diff
@@ -0,0 +1,11 @@
+package users
+
+import (
+	"context"
+
+	"github.com/ucl-arc-tre/portal/internal/types"
+)
+
+func (s *Service) InviteUser(ctx context.Context, email string, sponsor types.Sponsor) error {
```
t-young31, line 9: this should, in my opinion, be idempotent. so if a user has already been invited then they're not invited again (so persist in the database and check if they exist – if they don't then send invite and persist if successful)

web/cypress/e2e/pages/people.cy.ts
```diff
@@ -57,3 +57,20 @@ describe("Import approved researchers", () => {
...
+    cy.get("[data-cy='show-invite-input']").should("not.exist");
+  });
+
+  it("should be visable to & usable by an admin user", () => {
+    cy.loginAsAdmin();
+    cy.visit("/people");
+    cy.get("[data-cy='show-invite-input']").should("be.visible").click();
+
+    cy.get("input[name='email']").should("be.visible").type("lab.amanda.ch@gmail.com");
+    cy.get("[data-cy='send-invite']").should("be.visible").click();
```
t-young31, line 74: maybe we don't actually want to send invites in the e2e – I'd probably mock this response. @jhughes982 what do you think?

t-young31 (approved):

t-young31 (changes requested):
wrong button 🤦🏼

t-young31:
https://github.com/ucl-arc-tre/portal/pull/172#issuecomment-3079040787

internal/controller/entra/main.go
```diff
@@ -97,3 +98,27 @@ func (c *Controller) IsStaffMember(ctx context.Context, username types.Username)
...
+	requestBody := graphmodels.NewInvitation()
+	invitedUserEmailAddress := email
+	inviteRedirectUrl := config.EntraInviteRedirectURL()
+	sendInvitationMessage := true
+
+	requestBody.SetInvitedUserEmailAddress(&invitedUserEmailAddress)
+	requestBody.SetInviteRedirectUrl(&inviteRedirectUrl)
+	requestBody.SetSendInvitationMessage(&sendInvitationMessage)
+
+	message := "You have been invited to join the UCL ARC Portal by " + string(sponsor.ChosenName)
```
t-young31, line 112: potentially want a conditional on here if they've not set their chosen name then fallback to their username?

internal/controller/entra/main.go
```diff
@@ -97,3 +98,27 @@ func (c *Controller) IsStaffMember(ctx context.Context, username types.Username)
...
 }
+
+func (c *Controller) SendInvite(ctx context.Context, email string, sponsor types.Sponsor) error {
+	requestBody := graphmodels.NewInvitation()
+	invitedUserEmailAddress := email
+	inviteRedirectUrl := config.EntraInviteRedirectURL()
+	sendInvitationMessage := true
+
+	requestBody.SetInvitedUserEmailAddress(&invitedUserEmailAddress)
+	requestBody.SetInviteRedirectUrl(&inviteRedirectUrl)
```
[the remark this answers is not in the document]
acholyn: returns a 500

deploy/dev/config.sample.yaml
```diff
@@ -8,6 +8,7 @@ entra:
   tenant_id: <CHANGE_ME>
```
t-young31, line 8: could you update the comment above this line with the additional permissions needed for the invite to work

deploy/dev/config.sample.yaml
```diff
@@ -8,6 +8,7 @@ entra:
...
   client_id: <CHANGE_ME>
   client_secret: <CHANGE_ME>
+  invite_redirect_url: <CHANGE_ME> # where you want to send the user when they accept the invite, eg. localhost:8000/profile
```
t-young31, line 11: not a URL(?)
```suggestion
  invite_redirect_url: <CHANGE_ME> # where to send the user when they accept the invite, eg. http://localhost:8000/profile
```

internal/handler/web/users.go
```diff
@@ -89,3 +89,30 @@ func (h *Handler) PostUsersApprovedResearchersImportCsv(ctx *gin.Context) {
...
+
+	var invite openapi.PostUsersInviteJSONRequestBody
+	if err := bindJSONOrSetError(ctx, &invite); err != nil {
+		return
+	}
```
t-young31, line 109: nit: probably pointless optimisation (sorry!) – if the request is invalid then we make an unnecessary database call. I'd put the JSON bind at the top of the function

web/src/components/people/ExternalInvite.tsx
```diff
@@ -0,0 +1,76 @@
...
+    } catch (err) {
+      console.error("Invite post error:", err);
+    } finally {
+      setButtonDisabled(false);
+      setIsLoading(false);
+      setEmail("");
+      setShowSuccessMessage(true);
+      setTimeout(() => {
+        setShowSuccessMessage(false);
+      }, 1000);
```
jhughes982, line 40: I would make the success or failure more explicit to the user. My eyes scanned around the page after I hit submit and I missed the changing of the message.

web/src/components/people/ExternalInvite.tsx
```diff
@@ -0,0 +1,76 @@
...
+    try {
+      setButtonDisabled(true);
+      setIsLoading(true);
+      if (typeof email === "string") {
+        const response = await postUsersInvite({ body: { email } });
+        console.log("response", response);
+      } else {
+        // redundant maybe?
+        console.error("Email not in correct format.");
+      }
```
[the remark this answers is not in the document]
acholyn: the uikit + html does validation for email format on its own so I did think it might be redundant

jhughes982:
My initial thought here is that the people page is now doing three things simultaneously (listing researchers, inviting researchers, and importing researchers). This might be fine, but I guess my gut-feeling would be to keep the UX as simple as it can be. At the moment my eyes scan around at the different elements of the page and someone new to the portal might have too many things to digest at once. 

Perhaps even just some text at the top of the page could be helpful here to explain what the page is for?

internal/handler/web/users.go
```diff
@@ -89,3 +89,31 @@ func (h *Handler) PostUsersApprovedResearchersImportCsv(ctx *gin.Context) {
...
+
+	// todo: check if invitee already exists in our db
+
+	if err := h.users.InviteUser(ctx, invite.Email, sponsor); err != nil {
+		setError(ctx, err, "Failed to send invite")
+		return
+	}
+
+	ctx.Status(http.StatusNoContent)
```
jhughes982, line 118: This might be one more for @t-young31 . Does this need to pass some sort of result to the user? At the moment if a server error is caught the user won't be notified about it. So even if a `500` is returned the UI will still say `invitation sent` and it looks like the invite was sent ok.

I think the error would need to be handled at least on the frontend. But possibly also caught on the backend and sent as a `result`. BUt I'll defer to @t-young31 here for more guidance.

web/src/components/people/ExternalInvite.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const [email, setEmail] = useState("");
+  const [showSuccessMessage, setShowSuccessMessage] = useState(false);
+  const [showErrorMessage, setShowErrorMessage] = useState(false);
+  const [errorMessage, setErrorMessage] = useState("");
+
+  function handleShowDialog() {
+    setDialogVisible(true);
+  }
+
+  async function handleSumbit(e: React.FormEvent<HTMLFormElement>) {
```
jhughes982, line 22:
```suggestion
  async function handleSubmit(e: React.FormEvent<HTMLFormElement>) {
```

internal/handler/web/users.go
t-young31: I guess depends what we want to show the user. Just the response code could be enough – e.g. `204` -> 'Invite sent', `406` -> 'Sorry your request wasn't valid', `500` -> 'Sorry something went wrong. Please try again'

web/src/components/people/ExternalInvite.tsx
```diff
@@ -0,0 +1,114 @@
...
+  function handleShowDialog() {
+    setDialogVisible(true);
+  }
+
+  async function handleSubmit(e: React.FormEvent<HTMLFormElement>) {
+    e.preventDefault();
+    try {
+      setButtonDisabled(true);
+      setIsLoading(true);
+      if (typeof email === "string") {
```
t-young31, line 27: what's this type check for? I don't think it can be anything other than a string(?)

t-young31 (approved):
🚀

web/src/components/people/ExternalInvite.tsx
```diff
@@ -0,0 +1,111 @@
...
+import Dialog from "../ui/Dialog";
+
+export default function ExternalInvite() {
+  const [buttonDisabled, setButtonDisabled] = useState(false);
+  const [isDialogVisible, setDialogVisible] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+  const [email, setEmail] = useState("");
+  const [showSuccessMessage, setShowSuccessMessage] = useState(false);
+  const [showErrorMessage, setShowErrorMessage] = useState("");
+
```
jhughes982, line 16: The current naming of `showErrorMessage` indicates it's a boolean (i.e. it is either shown or not shown). Calling it something like `errorMessage` is a bit closer to what the variable actually represents.

PR #199, opened 2025-07-31
Has data studies column

### LINKED ISSUE #178: Understand purpose of "Got data" field in Study Form
It is currently a required field - but is very confusingly worded. Need to find out what it is used for and whether it's important to keep.

jhughes982:
## Resolves #178 

- This PR adds a 'has_existing_data' column to the database, indicating whether the study has data already or not.
- The frontend has also been updated with the new question

### Checklist

### DIFF
--- a/api.web.yaml
+++ b/api.web.yaml
@@ -598,6 +598,7 @@ components:
         - title
         - data_controller_organisation
         - additional_study_admin_usernames
+        - has_existing_data
       properties:
         title:
           type: string
@@ -610,6 +611,9 @@ components:
           items:
             type: string
           description: List of additional study administrator usernames (empty array if none)
+        has_existing_data:
+          type: boolean
+          description: Whether the study involves existing data (e.g., data already collected or available)
         data_controller_organisation:
           type: string
           description: The organisation acting as data controller for the study (e.g., "UCL" or custom organization name)
--- a/web/src/components/studies/CreateStudyForm.tsx
+++ b/web/src/components/studies/CreateStudyForm.tsx
@@ -29,7 +29,7 @@ export type StudyFormData = {
   description: string;
   owner: string;
   additionalStudyAdminUsernames: { value: string }[];
-  hasExistingData: boolean;
+  hasExistingData: string;
   dataControllerOrganisation: string;
   cagReference: number;
   dataProtectionPrefix: string;
@@ -317,7 +317,6 @@ export default function CreateStudyForm(CreateStudyProps: CreateStudyProps) {
                   value="true"
                   {...register("hasExistingData", {
                     required: "Please select whether your study involves existing data",
-                    setValueAs: (value) => value === "true",
                   })}
                 />
                 Yes - we have existing data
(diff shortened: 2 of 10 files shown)


api.web.yaml
```diff
@@ -610,6 +611,9 @@ components:
           items:
             type: string
           description: List of additional study administrator usernames (empty array if none)
+        has_existing_data:
+          type: boolean
+          description: Whether the study involves existing data (e.g., data already collected or available)
```
jhughes982, line 616: This could alternatively be a string if other kinds of answers are allowable

web/src/components/studies/Studies.tsx
```diff
@@ -51,6 +51,7 @@ export default function Studies(props: Props) {
         additional_study_admin_usernames: data.additionalStudyAdminUsernames
           .map((admin) => admin.value.trim())
           .map((username) => `${username}${domainName}`),
+        has_existing_data: data.hasExistingData === "true", // Convert to boolean
```
jhughes982, line 54: Since the radio button stores data asa string, it is converted here to make it a boolean. But we can think more about the data type if needed.

web/src/components/studies/CreateStudyForm.tsx
```diff
@@ -303,6 +304,44 @@ export default function CreateStudyForm(CreateStudyProps: CreateStudyProps) {
         </fieldset>
 
         {/* second step */}
+        <fieldset className={getFieldsetClass(2)}>
+          <legend>Existing data</legend>
+
+          <Label>
+            Does your study involve existing data? *
+            <div className={styles["existing-data-radio-group"]}>
```
jhughes982, line 312: Happy to be advised further on the exact wording of the question and answers.

api.web.yaml
acholyn: think a bool makes most sense cause either there is data or not, right?

web/src/components/studies/CreateStudyForm.tsx
finleybacon: > I suppose the asset's location is talking about where the asset is going to be once they create a Project and ingress their data. Since everyone will be completing the asset register before they even have an approved Study, let alone a Project.

maybe this needs further thought, I don't know if this has been explicitly agreed

web/src/components/studies/Studies.tsx
jhughes982: I considered both. In the end I though having two very explicit options to choose from would be more clear to the user. But happy to rethink this if there are other ideas?

web/src/components/studies/CreateStudyForm.tsx
jhughes982: Thanks all for the input. I'd like to reach a consensus before I start implementing this further. So we could:
1. Go with @finleybacon 's idea of having the follow-up question, and then store the data as a boolean in a column called `has_existing_insecure_data`.
2. Not ask the question at all and rely on the asset questionnaire.
3. Some other idea?

Would it be easier to discuss on a call?

PR #204, opened 2025-08-01
adding study approval column

### LINKED ISSUE #186: Add status field for study
**As a** IG admin
**I want** to have a status field of a study
**So that** a study can be approved and be eligible for project creation

## Possible values

 - Incomplete: Waiting for study owner to fill in data assets and contracts
 - Pending: Waiting for IG approval/rejection
 - Approved
 - Rejected

jhughes982:
## Resolves #186 

- This PR adds an `approval_status` column to the study submission.
- The status is set to `Incomplete` on initial study submission.

## Adding tests
- I think the study creation process is nearing completing at this point. So I will make a follow-up PR to add some tests.

### Checklist


internal/service/studies/main.go
```diff
@@ -185,6 +185,7 @@ func (s *Service) createStudy(user types.User, studyData openapi.StudyCreateRequ
 		OwnerUserID:                user.ID,
 		Title:                      studyData.Title,
 		DataControllerOrganisation: studyData.DataControllerOrganisation,
+		ApprovalStatus:             "Incomplete", // Initial status is "Incomplete" until the contract and assets are created
```
t-young31, line 188: let's use the enum value here https://github.com/ucl-arc-tre/portal/blob/59c175a0ab211b4993a4f8d980816072d198d9fd/internal/openapi/web/main.gen.go#L50

api.web.yaml
```diff
@@ -598,6 +598,7 @@ components:
         - title
         - data_controller_organisation
         - additional_study_admin_usernames
+        - approval_status
```
t-young31, line 601: I think this needs to be in the `Study` rather than `StudyBase` schema, as it's not settable in the request

t-young31: yeah - no idea. [upstream](https://github.com/oapi-codegen/oapi-codegen) weirdness

could consider splitting out the enum into its own `StudyApprovalStatus` schema for a (hopefully!) nicer generated name

t-young31 (approved):
LGTM

PR #205, opened 2025-08-01
tweaking no studies message

jhughes982:
## Works towards #5 

- A quick PR to improve the UX for non-staff members.
- They will now see a more clear message on the studies page telling them what the current status is when there are no studies added yet.
- For staff members, they will get their own message and the option to create a new study

Screenshot 2025-08-01 at 15 22 37

### Checklist


t-young31 (approved):
lovely

acholyn (approved):
me gusta

PR #215, opened 2025-08-04
fix base role assignment

### LINKED ISSUE #214: Persisted users should always be assigned a base role
## Description

**As a** dev
**I want to** have all users be assigned a base role
**So that** they can login to the portal

t-young31:
## Resolves #214

Found earlier with @acholyn  – whoops

### Checklist

### DIFF
--- a/internal/middleware/user.go
+++ b/internal/middleware/user.go
@@ -45,13 +45,13 @@ func (u *UserSetter) setUser(ctx *gin.Context) {
 		ctx.Set(userContextKey, user)
 		return
 	}
-	user, err := u.users.PersistedUser(username)
+	user, _, err := u.users.PersistedUser(username)
 	if err != nil {
 		log.Err(err).Msg("Failed to get user")
 		ctx.AbortWithStatus(http.StatusInternalServerError)
 		return
 	}
-	if hasRole, err := rbac.HasRole(user, rbac.Base); err != nil && !hasRole {
+	if hasBaseRole, err := rbac.HasRole(user, rbac.Base); err != nil && !hasBaseRole {
 		_, _ = rbac.AddRole(user, rbac.Base)
 	}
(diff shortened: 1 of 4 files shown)


internal/middleware/user.go
```diff
@@ -47,34 +45,14 @@ func (u *UserSetter) setUser(ctx *gin.Context) {
...
-	result := u.db.Where("username = ?", username).
-		Attrs(types.User{
-			Username: username,
-			Model:    types.Model{CreatedAt: time.Now()},
-		}).
-		FirstOrCreate(&user)
-	if result.Error != nil {
-		return user, types.NewErrServerError(result.Error)
+	if hasRole, err := rbac.HasRole(user, rbac.Base); err != nil && !hasRole {
+		_, _ = rbac.AddRole(user, rbac.Base)
```
acholyn, line 55: what situations are we expecting this to proc in? (since in PersistedUser if the user is created they should then be given base role)

t-young31: in the case the user has already been added with the wrong function, then tries to log in 🤦🏼

acholyn (approved):
🙌 😌

internal/service/users/user.go
```diff
@@ -54,6 +55,28 @@ func (s *Service) AllUsers() ([]openapi.UserData, error) {
...
+	if result.Error != nil {
+		return user, types.NewErrServerError(result.Error)
+	}
+	userWasCreated := result.RowsAffected > 0
+	if userWasCreated {
+		if _, err := rbac.AddRole(user, rbac.Base); err != nil {
+			return user, fmt.Errorf("failed assign user base role: %v", err)
+		}
+	}
+	return user, nil
```
acholyn, line 77: can we also return userWasCreated so it can be used in other conditionals after 🙏

PR #218, opened 2025-08-05
Profile steps refactor

jhughes982:
## Works towards #6 

- This PR refactors the components involved in completing the profile steps
- The steps have now been abstracted into a reusable component and the logic has been slightly rearranged to make things cleaner
- I also added a animated arrow to make it clear to the user to scroll down to complete the current step
- This is in preparation for reusing the steps UI and UX for the study management.

### Checklist

### DIFF
--- a/web/src/components/ui/steps/StepProgress.module.css
+++ b/web/src/components/ui/steps/StepProgress.module.css
@@ -1,5 +1,5 @@
 .step-progress-container {
-  width: 60%;
+  width: 80%;
   margin: 0 auto;
   padding: 1.5rem;
   border: 1px solid #e5e7eb;
--- a/web/src/components/profile/ProfileSummaryCard.module.css
+++ b/web/src/components/profile/ProfileSummaryCard.module.css
@@ -1,6 +1,5 @@
 .profile-summary-container {
-  width: 70%;
-  background: var(--background);
+  width: 80%;
   border: 1px solid #e5e7eb;
   border-radius: 8px;
   padding: 1.5rem;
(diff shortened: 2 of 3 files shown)


web/src/types/steps.d.ts
```diff
@@ -0,0 +1,18 @@
...
+type StepProgressProps = {
+  steps: Step[];
+  isComplete?: boolean;
+  completionTitle?: string;
+  completionSubtitle?: string;
+  completionButtonText?: string;
+  completionButtonHref?: string;
+  introText?: string;
+  ariaLabel?: string;
+};
```
jhughes982, line 18: These will be reused by the study management tasks, so I extracted them out here.

web/src/components/ui/steps/StepProgress.module.css
```diff
@@ -1,8 +1,15 @@
...
+  border: 1px solid #e5e7eb;
+  border-radius: 0.5rem;
+}
+
 .intro-text {
-  margin: 0 0 1.5rem 0;
-  font-size: 0.875rem;
-  color: #6b7280;
-  line-height: 1.6;
+  margin: 0 0 1.2rem 0;
```
acholyn, line 10: might be personal preference but I feel like this + the bottom padding on completion-header makes it look quite far from the line, I'd pick one rather than both

t-young31:
@acholyn want to have the casting vote?

acholyn:
I like the new flow but I do like the wider width for the part after the arrow 😅 
tangentally, the first box isn't centred on safari but that can be fixed later

jhughes982:
> I like the new flow but I do like the wider width for the part after the arrow 😅 tangentally, the first box isn't centred on safari but that can be fixed later

It looks like a few different elements are not positioned correctly on Safari. I'll make a ticket for this to follow up separately.

t-young31:
I don't see the commit in the PR @jhughes982 👀

jhughes982:
> I don't see the commit in the PR @jhughes982 👀

Oops, forgot to hit push 😅

t-young31 (approved):
🚢

PR #221, opened 2025-08-06
Amanda/logout

### LINKED ISSUE #219: adding logout functionality
Would be nice to log out
Helpful for devs swapping between accounts but also good op sec

acholyn:
### Checklist
- adds a button to the bottom of the nav (hard to add into Header component nicely but if overwhelming preference for that then I can have another go)
- request should go to oauth2 signout first then entra and then redirect to the profile page (reusing the invite redirect url to reduce config but could change this to just be to home and tweak the invite usage)


jhughes982:
Didn't we discuss this before and decide that it wasn't needed? 

Not that I feel strongly about it either way, this might be a nice user feature? But we should aim to be aligned in the project priorities.

jhughes982:
> Didn't we discuss this before and decide that it wasn't needed?
> 
> Not that I feel strongly about it either way, this might be a nice user feature? But we should aim to be aligned in the project priorities.

Oh, I just noticed an issue was added for this. Still, it would be good in general to discuss as a team and decide what the priority of the feature is. That way we avoid feature drift.

internal/handler/web/users.go
```diff
@@ -126,3 +126,10 @@ func (h *Handler) PostUsersInvite(ctx *gin.Context) {
 
 	ctx.Status(http.StatusNoContent)
 }
+
+func (h *Handler) GetLogout(ctx *gin.Context) {
```
t-young31, line 130: should probably be in `internal/handler/web/main.go`

web/src/components/nav/Nav.module.css
```diff
@@ -60,3 +60,21 @@
...
   }
 }
+
+.logout {
+  position: absolute !important;
+  width: 150px;
+  bottom: 10vh;
+  align-self: end;
+  font-weight: 400 !important;
+  color: var(--green-200) !important;
```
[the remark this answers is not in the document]
acholyn: unfortunately the styling for those attributes doesn't take without `!important` because of how the UIkit works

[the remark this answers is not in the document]
jhughes982: Thats a shame.

internal/handler/web/main.go
```diff
@@ -24,3 +28,10 @@ func New() *Handler {
 		auth:       auth.New(),
 	}
 }
+
+func (h *Handler) GetLogout(ctx *gin.Context) {
+
+	logoutUrl := "/oauth2/sign_out?rd=https://login.microsoftonline.com/" + config.EntraCredentials().TenantID + "/oauth2/v2.0/logout?post_logout_redirect_uri=" + config.EntraInviteRedirectURL()
+
+	ctx.Redirect(http.StatusFound, logoutUrl)
+}
```
jhughes982, line 37: Could this equally work as a direct link from the frontend? For login we currently have:

```
      <Button className={styles.button} size="large" href={`/oauth2/start?`} cy="login">
        Login with UCL SSO
      </Button>
```

I don't feel strongly about it one way or the other. Unless the implementation is simpler via a direct link?

jhughes982: I see, it would be nice if there was a way to utilise something like `href={`/oauth2/logout`}` but fair enough if it's not possible.

Issue #109, opened 2025-06-03
Study & project member view


acholyn:
**As a** study owner or administrator (IAO/IAA)
**I want** to be able to view all the people which belong to my projects
**So that** I can see (but not edit) their training and agreement status
 
## Acceptance criteria
- [ ] IAO/IAA can access `/people` GET with read only
- [ ] `/people` GET returns users associated with the studies via the projects
  -  i.e `studies.projects.users` WHERE `IAO/IAA username` === `user.username` 
  - [ ] card display of user list (following updates in #788, use components ) - include IAO & IAA in list
 - [ ] option to organise by username, study name and project name

## Dependencies

 - [x] #11 
 - [x] #788

jhughes982:
Just to clarify the ticket and make sure I'm following this correctly: the title is currently `Approved researcher people view` but the ticket description refers to information asset owners. Is the title just out of date? Should it be more like: `view people belonging to projects`?

acholyn:
have updated, hopefully more clear now? you get the people from projects and from the study (in case of iao and iaa)

jhughes982:
> have updated, hopefully more clear now? you get the people from projects and from the study (in case of iao and iaa)

So is the ticket about viewing people belonging to both studies AND projects? So far I've been following this from the ticket description:

> I want to be able to view all the people which belong to my projects

But is it actually: `I want to be able to view all the people who belong to my studies and projects`?

If so, are these separate views? Or one big list? Also, at the moment there is no concept of a `study member` right unless you mean the study admin?

acholyn:
it is for IAOs and IAAs to see all the people associated with the studies they own/manage. Most people are associated with the study through a project since we only have one IAO and one IAA per study

>Also, at the moment there is no concept of a study member right unless you mean the study admin?

I would consider members of projects within a study to be that study's members since the study is the parent entity

jhughes982:
For fetching people belonging to a study/project, do we want RBAC to fully enforce access, or do we give a blanket `/users/:id read` policy and restrict access via SQL?

acholyn:
I _think_ rbac, where only people with read access can read members? but also not entirely convinced I appreciate the difference in approaches @t-young31 ?

jhughes982:
> it is for IAOs and IAAs to see all the people associated with the studies they own/manage. Most people are associated with the study through a project since we only have one IAO and one IAA per study
> 
> > Also, at the moment there is no concept of a study member right unless you mean the study admin?
> 
[...]

PR #229, opened 2025-08-11
enabling approved researchers to invite externals

acholyn:
## Related to #3 
also related to #109 

### Checklist


finleybacon:
Yes definitely need to be a full member of staff - and actually I had been working on the assumption that they also need to be Study Owners (IAO/IAAs). In the Study Agreement they explicitly agree to a section on [Safe Researchers](https://isms.arc.ucl.ac.uk/agreement_templates/rism15-study_agreement/#safe-researchers).

I think it sounds fair to enable it only for Study owners initially, and potentially widen access to all Staff if there's demand? But start more restrictive

PR #240, opened 2025-08-13
222/ops role initial

acholyn:
### Checklist
- added new `TreOpsStaff` role that has read perms on everything
- added field to config
- updating nav perms
- treops people view
- function to get users by given role

drive by:
- made `.role` a global class
- refactored people page users table into reusable component `UserDataTable`
- refactored looping through users to get details into a new function
- `UserById` now takes uuid
- new `UsersByIds` function


internal/rbac/main.go
```diff
@@ -114,6 +115,22 @@ func HasRole(user types.User, role RoleName) (bool, error) {
 	return hasRole, types.NewErrServerError(err)
 }
 
+func GetUsersWithRole(role RoleName) ([]types.User, error) {
```
t-young31, line 118: I think this should be
```go
func GetUserIdsWithRole(role RoleName) ([]uuid.UUID, error) {
```
as, as a consumer I'd expect the `User`s returned would have the username set (and creation time etc.)

acholyn: the casbin method returns a slice of strings and the `UserById` service takes in a string, so can I just return strings or is it worth converting them to the uuid type? 🤔

web/cypress/e2e/pages/people.cy.ts
```diff
@@ -30,6 +30,18 @@ describe(`People page content`, () => {
...
+  it("should show content for TRE ops staff", () => {
+    cy.loginAsBase();
+    cy.mockAuthAsTreOpsStaff();
+    cy.visit("/people");
+
+    cy.contains("You do not have permission to view this page").should("not.exist");
+    cy.contains("View approved researchers").should("be.visible");
+
+    // should only show approved researchers...best way to test this? Or not bother here?
+    // idea would be to create 2 users, one AR and only AR is in the response
```
acholyn, line 43: ideas pls

internal/service/users/user.go
```diff
@@ -21,7 +21,46 @@ func (s *Service) AllUsers() ([]openapi.UserData, error) {
...
+
+	// get all approved researcher users from db
+	userIds, err := rbac.GetUserIdsWithRole(rbac.ApprovedResearcher)
+	if err != nil {
+		return usersData, types.NewErrServerError(err)
+	}
+
+	users := []types.User{}
+	for _, userId := range userIds {
+		user, err := s.UserById(userId)
```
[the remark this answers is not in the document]
t-young31: cf. https://github.com/ucl-arc-tre/portal/blob/4ec8e3b2167b7cc48e9ee4c2cdf3247029c1f422/internal/service/studies/main.go#L154

internal/rbac/init.go
```diff
@@ -89,3 +101,25 @@ func persistedAdminUsers() []types.User {
 	}
 	return users
 }
+
+func persistedTreOpsStaffUsers() []types.User {
```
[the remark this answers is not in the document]
acholyn: will follow up with #241

internal/rbac/main.go
```diff
@@ -114,6 +115,24 @@ func HasRole(user types.User, role RoleName) (bool, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	ids := []uuid.UUID{}
+	for _, id := range userIds {
+		uid, err := uuid.Parse(id)
+		if err != nil {
+			panic(err)
+		}
```
t-young31, line 129:
```suggestion
		uid := uuid.MustParse(id)
```

internal/service/users/user.go
```diff
@@ -21,7 +21,41 @@ func (s *Service) AllUsers() ([]openapi.UserData, error) {
 		return usersData, types.NewErrServerError(result.Error)
 	}
 
-	// then loop through each and get their agreements & roles
+	usersData, err := s.usersData(users)
+	if err != nil {
+		return usersData, err
+	}
+
+	return usersData, nil
```
t-young31, line 29:
```suggestion
	return s.usersData(users)
```

t-young31 (approved):
LGTM – will leave for @jhughes982 to look at the frontend

web/src/pages/people/index.tsx
```diff
@@ -26,12 +28,14 @@ export default function PeoplePage() {
...
-              : "You do not have permission to view this page"
+            : isTreOpsStaff
+              ? "View approved researchers"
+              : isIAO
+                ? "View users in your projects or invite a collaborator"
+                : "You do not have permission to view this page"
         }
       />
-      {isAdmin ? <AdminView /> : isIAO && <IAOView />}
+      {isAdmin ? <AdminView /> : isTreOpsStaff ? <TreOpsStaffView /> : isIAO && <IAOView />}
```
jhughes982, line 38: It looks like these `View` files share some basic common functionality. I wonder if it would be more efficient to fetch the people in `pages/people/index.tsx` and then pass them to all subsequent components. Then you could potentially get rid of all the separate `View` files and have one file that handles the UI based on the different roles.

The danger would be that we start making all these different people `View` files and end up with a lot of files and repeated logic.

web/src/components/people/AdminView.tsx
```diff
@@ -43,113 +27,16 @@ export default function AdminView() {
...
-      setUsers(updatedPeople);
-    }
-  };
-
-  const handleEditTrainingClick = (id: string) => {
-    setId(id);
-    setTrainingDialogOpen(true);
-  };
-
   if (!users) return null;
```
jhughes982, line 31: This should have more context. The data fetch should handle errors and distinguish an error from the scenario where there are genuinely no users. If there are no users then the page should say something like "no users found"

web/src/components/people/UserDataTable.tsx
```diff
@@ -0,0 +1,134 @@
...
+  users: Array<UserData>;
+  setUsers: (users: Array<UserData>) => void;
+  isLoading: boolean;
+};
+export default function UserDataTable(Props: Props) {
+  const { canEdit, users, setUsers, isLoading } = Props;
+  const [trainingDialogOpen, setTrainingDialogOpen] = useState(false);
+
+  const [id, setId] = useState("");
+
```
jhughes982, line 21: Is this a userId? if so I'd call it that.

web/src/components/people/UserDataTable.tsx
```diff
@@ -0,0 +1,134 @@
...
+        <Loading message="Loading users..." />
+      </div>
+    );
+  }
+
+  return (
+    <>
+      {trainingDialogOpen && (
+        <TrainingForm id={id} setTrainingDialogOpen={setTrainingDialogOpen} updatePersonUI={updatePersonUI} />
+      )}
```
jhughes982, line 62: Again, if this is a userId i'd be more explicit in the naming.

web/src/pages/people/index.tsx
acholyn: I also thought about that and wondered if it might be problematic for someone to be able to technically have access to all the users (in the network response) but they only see a subset in the UI. I opted to air on the side of caution but happy to have a conversation about it. It may also be that at _this_ stage they're fairly similar but will grow more different over time

acholyn: there's also the new UserDataTable which is the bulk of the similarity, and then it's really only the fetching that's split out for each view

web/src/components/people/UserDataTable.tsx
```diff
@@ -0,0 +1,134 @@
...
+      person.training_record.training_records.push(training);
+      setUsers(updatedPeople);
+    }
+  };
+  const handleEditTrainingClick = (id: string) => {
+    setId(id);
+    setTrainingDialogOpen(true);
+  };
+
+  if (!users) return null;
```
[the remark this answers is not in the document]
jhughes982: Did you decide not to implement the suggestion in the end (I noticed you marked it as resolved)? 😅

web/src/pages/people/index.tsx
jhughes982: I'd probably just do it in this PR since the work is directly relevant to what the PR is about. Unless you felt strongly about it?

web/src/components/people/UserDataTable.tsx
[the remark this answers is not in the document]
acholyn: It's been handled in the separate views so it would've been duplicated if done on the table as well but will be reworked anyway through consolidation (above comment about removing views)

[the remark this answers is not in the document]
jhughes982: Sounds good. Just for future reference, it would be good to leave a brief comment before you resolve the conversations just so that I know what the decision has been.

As a side note, if you have already done a check for empty users in the parent component I don't think you need to repeat it here.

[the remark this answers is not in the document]
acholyn: 🫡 wilco

web/src/pages/people/index.tsx
```diff
@@ -26,12 +64,33 @@ export default function PeoplePage() {
...
+          <div className={styles["no-users-found"]}>No users found</div>
+          {errorMessage && (
+            <Alert type="error">
+              <AlertMessage>{errorMessage}</AlertMessage>
+            </Alert>
+          )}
+        </Box>
+      ) : (
+        <UserDataTable canEdit={true} users={users} setUsers={setUsers} isLoading={isLoading} />
+      )}
```
jhughes982, line 93: Is there a benefit to passing `isLoading` to the child component here? You could equally render the loading message in this file and save rendering a component when it is not ready yet.

jhughes982 (approved):
Looks good.

PR #251, opened 2025-09-01
Simple version of search

acholyn:
### Checklist

- adds api and logic for search on people page
- actual implementation on frontend tbd; blocked until responses from DSH
- simple search bar implementation on people page

you can test with eg. `http://localhost:8000/web/api/v0/users?find=portal`

Drive by:
- hiding projects in nav
- pseudo-caching users on people page? see https://github.com/ucl-arc-tre/portal/pull/251/commits/c77d50dde09130fa6b7ac48a21d27bffd6a2cadc

#### Limitations
- even when trying `contains` instead of `startswith` for the entra query, I couldn't find users for "admin" or "ahl" searches


api.web.yaml
```diff
@@ -144,6 +144,13 @@ paths:
   /users:
     get:
       description: Get all the users a user has access to view
+      parameters:
+        - in: query
+          name: find
+          required: false
+          description: user details to lookup by
```
t-young31, line 151: better description and some examples here are probably warranted

internal/controller/entra/main.go
```diff
@@ -184,3 +185,44 @@ func (c *Controller) AddtoInvitedUserGroup(ctx context.Context, email string) er
 	return err
 
 }
+
+func (c *Controller) SearchForUser(ctx context.Context, query string) ([]string, error) {
```
t-young31, line 189:
```go
// SearchUsernames searches entra for usernames given a query of email, user principal or display name
func (c *Controller) FindUsernames(ctx context.Context, query string) ([]types.Username, error) {
```

internal/controller/entra/main.go
```diff
@@ -184,3 +185,44 @@ func (c *Controller) AddtoInvitedUserGroup(ctx context.Context, email string) er
...
+	// do filter query and check in email, principal name and display name
+
+	filterQuery := fmt.Sprintf(
+		`startswith(displayName,'%s') or startswith(userPrincipalName,'%s') or startswith(givenName,'%s') or startswith(mail,'%s')`,
+		query, query, query, query,
```
t-young31, line 194: I know it's not sql but this makes me scared about injection. I think the query should be validated to match something like `^\w[\w.\s0-9@]+\w$` (suggestions welcome)

api.web.yaml
acholyn: https://github.com/ucl-arc-tre/portal/pull/251/commits/58c314f75f8821f74fb6d94cad5ed6f13fd5d1e2

internal/controller/entra/main.go
acholyn: presumably earlier is better for validation? so would do it in `GetUsers`?

t-young31: I'd actually do it in this function – so all consumers get a `InvalidObjectError` if the query is invalid

internal/handler/web/users.go
```diff
@@ -16,7 +17,8 @@ import (
 	"github.com/ucl-arc-tre/portal/internal/rbac"
 )
 
-func (h *Handler) GetUsers(ctx *gin.Context) {
+func (h *Handler) GetUsers(ctx *gin.Context, _ openapi.GetUsersParams) {
```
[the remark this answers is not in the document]
t-young31: yep 🙌🏼.  could rm the `query` var and check on `params.Find` – more understandable to me at least

internal/service/users/user.go
```diff
@@ -125,3 +125,25 @@ func (s *Service) findUser(user *types.User) (*types.User, error) {
...
+func (s *Service) SearchEntraForUsersAndMatch(ctx context.Context, query string) ([]openapi.UserData, error) {
+	// query entra
+	foundEmails, err := s.entra.SearchForUser(ctx, query)
+	if err != nil {
+		return nil, err
+	}
+	// then match to users in our db
+	users := []types.User{}
+
+	result := s.db.Where("username IN (?)", foundEmails).Find(&users)
```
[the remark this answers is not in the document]
t-young31: cool – probably worth checking

internal/controller/entra/main.go
```diff
@@ -184,3 +185,44 @@ func (c *Controller) AddtoInvitedUserGroup(ctx context.Context, email string) er
...
+	headers := abstractions.NewRequestHeaders()
+
+	configuration := &graphusers.UsersRequestBuilderGetRequestConfiguration{
+		Headers: headers,
+		QueryParameters: &graphusers.UsersRequestBuilderGetQueryParameters{
+			Filter: &filterQuery,
+		},
+	}
+
+	data, err := c.client.Users().Get(ctx, configuration)
```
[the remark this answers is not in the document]
acholyn: just returns all matches, no limits 😬 
Think we might be able to tack on a paginate func

internal/handler/web/users.go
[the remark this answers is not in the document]
acholyn: https://github.com/ucl-arc-tre/portal/pull/251/commits/03102b6fbd31fb785921eace0cd4235c51b478dd acceptabubble?

internal/service/users/user.go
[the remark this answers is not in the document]
acholyn: yeah, just returns an empty slice, no errors

internal/controller/entra/main.go
```diff
@@ -184,3 +186,47 @@ func (c *Controller) AddtoInvitedUserGroup(ctx context.Context, email string) er
 	return err
 
 }
+
+// FindUsernames searches entra for usernames given a query of email, user principal or display name
+func (c *Controller) FindUsernames(ctx context.Context, query string) ([]types.Username, error) {
+	queryRegex := regexp.MustCompile(`^\w[\w.\s0-9@]+\w$`)
+	queryIsValid := queryRegex.MatchString(query)
+	if !queryIsValid {
+		return nil, types.NewErrInvalidObject("invalid query")
```
t-young31, line 195:
```suggestion
		return nil, types.NewErrInvalidObject(fmt.Sprintf("invalid query [%v]", query))
```

internal/controller/entra/main.go
```diff
@@ -184,3 +186,47 @@ func (c *Controller) AddtoInvitedUserGroup(ctx context.Context, email string) er
...
+		Headers: headers,
+		QueryParameters: &graphusers.UsersRequestBuilderGetQueryParameters{
+			Filter: &filterQuery,
+			Top:    &resultLimit,
+		},
+	}
+
+	data, err := c.client.Users().Get(ctx, configuration)
+	if err != nil {
+		return nil, err
```
t-young31, line 216:
```suggestion
		return nil, types.NewErrServerError(err)
```

internal/controller/entra/main.go
```diff
@@ -184,3 +186,47 @@ func (c *Controller) AddtoInvitedUserGroup(ctx context.Context, email string) er
...
+	}
+
+	userMatches := data.GetValue()
+	if userMatches == nil {
+		return nil, types.NewNotFoundError("no users matching query found in tenant")
+	} else {
+	}
```
t-young31, line 223:
```suggestion
	if userMatches == nil || len(userMatches) == 0 {
		return nil, types.NewNotFoundError("no users matching query found in tenant")
	}
```

t-young31: question here about how we want to handle empty responses – `404` or `200` with empty slice?

web/src/pages/people/index.tsx
```diff
@@ -45,6 +55,30 @@ export default function PeoplePage() {
...
+      setUsers(originalUsers);
+    }
+  };
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setSearchTerm(event.target.value);
+    if (event.target.value === "") {
+      setUsers(originalUsers);
+    }
+  };
```
jhughes982, line 76: Its not essential (more of a nice to have) but it might be more efficient to attach the search term to a `useRef`. That way, the state isn't being updated with every keystroke and is just used as an when it is needed. 

I don't feel strongly about implementing it in this alternative way. But thought you might be interested to know about it.

web/src/pages/people/index.tsx
```diff
@@ -45,6 +55,30 @@ export default function PeoplePage() {
...
+  const handleUserSearchSubmit = (e: React.MouseEvent<HTMLButtonElement>) => {
+    e.preventDefault();
+    handleUserSearch(searchTerm);
+  };
```
jhughes982, line 80: Is this additional abstraction needed? Couldn't the logic all go in `handleUserSearch`?

web/src/pages/people/index.tsx
```diff
@@ -72,16 +106,40 @@ export default function PeoplePage() {
...
+              icon={<SearchIcon />}
+              onClick={handleUserSearchSubmit}
+              type="submit"
+              aria-label="submit user search query"
+            ></Button>
+            {searchErrorMessage && (
+              <Alert type="error">
+                <AlertMessage>{errorMessage}</AlertMessage>
+              </Alert>
+            )}
```
jhughes982, line 131: When I typed in something that didn't work (e.g. 'jh') I got a red cross on the search bar, but there was no accompanying error message to tell me what I did wrong. Also, if I then subsequently made a new search for something that did work (e.g. 'portal') the red cross never goes away.

web/src/pages/people/index.tsx
```diff
@@ -45,6 +55,30 @@ export default function PeoplePage() {
...
+      const response = await getUsers({ query: { find: query } });
+      if (response.response.ok && response.data) {
+        setUsers(response.data);
+      } else {
+        setSearchErrorMessage("Oops, something went wrong with your search. Refresh and try again");
+      }
+    } else {
+      setUsers(originalUsers);
+    }
+  };
```
jhughes982, line 69: Personal preference: the embedded if statements can be simplified by returning early e.g.

```
if (query === "") {
  setUsers(originalUsers);
  return;
}
...
```

internal/service/users/user.go
```diff
@@ -125,3 +125,20 @@ func (s *Service) findUser(user *types.User) (*types.User, error) {
...
 }
+
+func (s *Service) SearchEntraForUsersAndMatch(ctx context.Context, query string) ([]openapi.UserData, error) {
+	// query entra
+	usernames, err := s.entra.FindUsernames(ctx, query)
+	if err != nil {
+		return nil, err
+	}
+	// then match to users in our db
+	users := []types.User{}
```
jhughes982, line 136: How come we need to search entra first?

web/src/pages/people/index.tsx
```diff
@@ -72,16 +106,40 @@ export default function PeoplePage() {
...
+        {(isAdmin || isTreOpsStaff) && (
+          <form className={styles["search-container"]}>
+            <Input
+              placeholder="search users..."
+              id={styles.search}
+              name="search"
+              onChange={handleInputChange}
+              value={searchTerm}
+              aria-label="search users of the portal"
+            ></Input>
```
jhughes982, line 119: Would some help text here be beneficial so the user knows what terms they can search by?

jhughes982 (commented):
A few initial comments. Some of them might not be valid as I learn more about how this is intended to work 👍

internal/service/users/user.go
acholyn: we don't store the email and sometimes the userprincipalname is in the EXT format, so we'll have to search entra anyway for each query, it just seemed to make more sense to search entra and then match it to our records. The original/alt way would've been searching our db, if nothing then searching entra anyway and then matching to our db.
Also users _shouldn't_ exist on the portal without existing in the tenant (laura@example is the exception)

web/src/pages/people/index.tsx
acholyn: https://github.com/ucl-arc-tre/portal/pull/251/commits/045cc2464802aba82c28db257ec209f4a65cd326

web/src/pages/people/index.tsx
acholyn: https://github.com/ucl-arc-tre/portal/pull/251/commits/045cc2464802aba82c28db257ec209f4a65cd326

web/src/pages/people/index.tsx
acholyn: https://github.com/ucl-arc-tre/portal/pull/251/commits/045cc2464802aba82c28db257ec209f4a65cd326

shouldn't get the error now, only came up if the term was shorter than 3 characters, so probably an entra limitation

web/src/pages/people/index.tsx
acholyn: putting the code into the `handleUserSearch` function and then calling it from the form or button makes things unhappy due to expected types of events, I think it's cleaner and easier to read as separate things but if you have a code suggestion I'm open

web/src/pages/people/index.tsx
```diff
@@ -44,6 +54,33 @@ export default function PeoplePage() {
...
+        const response = await getUsers({ query: { find: query } });
+        if (response.response.ok && response.data) {
+          setUsers(response.data);
+          setSearchErrorMessage("");
+        } else {
+          setSearchErrorMessage("Oops, something went wrong with your search. Refresh and try again");
+        }
+      }
+    }
+  };
```
jhughes982, line 75:
```suggestion
 const handleUserSearch = async (query: string): Promise<void> => {
  if (query.length < 3) {
    setSearchErrorMessage("minimum 3 characters required to perform search");
    return;
  }

  if (query === "") {
    setUsers(originalUsers);
    return;
  }

  const response = await getUsers({ query: { find: query } });

  if (!response.response.ok || !response.data) {
    setSearchErrorMessage("Oops, something went wrong with your search. Refresh and try again");
    return;
  }

  setUsers(response.data);
  setSearchErrorMessage("");
};

```

I find nested `if` blocks hard to read. Using early returns makes the logic easier to follow, at least for me.

internal/service/users/user.go
t-young31: I'm pro search only! but will need some user feedback before we decide

acholyn: but also, presently I think entra does actually [limit](https://github.com/ucl-arc-tre/portal/blob/4e517bcc1c8631ffa7957399796d1f6b1a2599a8/internal/controller/entra/main.go#L202C5-L202C5) the page length to 20, we're just not utilising it on the frontend

web/src/pages/people/index.tsx
acholyn: 🤦‍♀️ 
forgot we could do it that way

internal/service/users/user.go
jhughes982: Yeah, that all makes sense. Happy to go with this for now and refine further once we get feedback 👍

web/src/pages/people/index.tsx
```diff
@@ -44,6 +54,40 @@ export default function PeoplePage() {
...
+      setUsers(originalUsers);
+      return;
+    }
+
+    const regex = `^\w[\w.\s0-9@]+\w$`;
+    const isValid = new RegExp(regex).test(query);
+    if (!isValid) {
+      setSearchErrorMessage("invalid query");
+      return;
+    }
```
jhughes982, line 75: I think the regex makes the length and empty string checks redundant right?

acholyn: ah yes, but now I'm getting page reloads when submitting, which isn't ideal 🤔

jhughes982: It's because of the [prevent default](https://github.com/ucl-arc-tre/portal/pull/251#discussion_r2318508114)

acholyn: grazie
https://github.com/ucl-arc-tre/portal/pull/251/commits/231331464a434482fc8556637c94c2d3d32da4e1

jhughes982: It looks like I am always hitting the invalid query error no matter what i type in now

web/src/pages/people/index.tsx
```diff
@@ -44,6 +54,33 @@ export default function PeoplePage() {
...
+  const canSearch = isTreOpsStaff || isAdmin;
+
+  const handleUserSearch = async (query: string) => {
+    const regex = /^\w[\w.\s0-9@]+\w$/;
+    const isValid = new RegExp(regex).test(query);
+    console.log(query, isValid);
+
+    if (!isValid) {
+      setSearchErrorMessage("invalid query; note that there's a minimum of 3 characters required to perform search");
+      setUsers(originalUsers);
```
jhughes982, line 66: I'm not entirely sure about setting the original users on an invalid query. Typical users (who are not the dev's) won't know that this is actually setting the originally fetched users and not some other set of filtered users. It's obvious with just a few users in the list, but as the list of users grows it may become confusing.

Perhaps it would be better to not return any users for an invalid query so that it is obvious that nobody was returned? I think instinctively people will modify their search term if something went wrong, or delete it to return to the default set of fetched users.

Happy to hear other thoughts though.

web/src/pages/people/index.tsx
```diff
@@ -44,6 +54,33 @@ export default function PeoplePage() {
   const isAdmin = userData?.roles.includes("admin");
   const isIAO = userData?.roles.includes("information-asset-owner");
   const isTreOpsStaff = userData?.roles.includes("tre-ops-staff");
+  const canSearch = isTreOpsStaff || isAdmin;
+
+  const handleUserSearch = async (query: string) => {
+    const regex = /^\w[\w.\s0-9@]+\w$/;
+    const isValid = new RegExp(regex).test(query);
+    console.log(query, isValid);
```
jhughes982, line 62: I'd remove the console.log.

web/src/pages/people/index.tsx
acholyn: If there are no results, it _should_ say "no results for <query>". Possibly we've changed something today that's made that not so. if it's invalid - len 0, doesn't match the regex, then they get an error

acholyn: image
image

jhughes982: Yes but if it hits an invalid regex the logic currently does: `setUsers(originalUsers);`. I'm suggesting that we don't do that and instead return no users for an invalid query.

But I could be wrong about that. I don't feel too strongly.

acholyn: so only if the query is empty again would we return the list of users?

jhughes982: Alternatively, if it hits an invalid query the users simply are not reset and it remains as it was for the most recent query?

jhughes982:
If I search for something that is valid but returns no results (e.g. 'test'), I see the correct "no users found for test" message.

But then if I start modifying the search term, the no users found message gets modified along with it which produces a message that shouldn't be possible (see screenshot below).

Screenshot 2025-09-03 at 12 11 09

jhughes982 (approved):
Looks good

acholyn:
image

PR #272, opened 2025-09-04
Invitation notification for existing tenant users

### LINKED ISSUE #269: invites broken
Presently, invites are broken because of a userData check that will throw an error and return if the user doesn't exist

acholyn:
### Checklist
- custom email notification if user already exists in tenant. In dev this is a little more fiddly since we're using 2 different tenants (mail from UCL, users from TRE)
- resolves #269


internal/controller/entra/main.go
```diff
@@ -134,41 +143,59 @@ func (c *Controller) IsStaffMember(ctx context.Context, username types.Username)
 
 func (c *Controller) SendInvite(ctx context.Context, email string, sponsor types.Sponsor) error {
```
t-young31, line 144: this function is too complex – I think should be split on the switch if the users exists in entra or not. e.g.

```go
func (c *Controller) SendInvite(ctx context.Context, email string, sponsor types.Sponsor) error {
	user, err := c.userData(ctx, types.Username(email))
	if err == nil {
		return c.sendInviteExistingEntraUser(...)
	} else if errors.Is(err, types.ErrNotFound) {
		return c.sendInviteNewEntraUser(...)
	} else {
		return err
	}
}
```

internal/controller/entra/main.go
```diff
@@ -134,41 +143,59 @@ func (c *Controller) IsStaffMember(ctx context.Context, username types.Username)
...
 
-	// check if user exists in entra, if yes, don't send invite
+	// check if user exists in entra, if not, send an invite; if yes, custom notification
 	user, err := c.userData(ctx, types.Username(email))
-	if err != nil {
+	if err != nil && strings.Contains(err.Error(), "does not exist") {
```
t-young31, line 148: this error is wrapped already so we can do
```suggestion
	if err != nil && errors.Is(err, types.ErrNotFound) {
```
and only have to do the string match in one place

internal/controller/entra/main.go
```diff
@@ -134,41 +143,59 @@ func (c *Controller) IsStaffMember(ctx context.Context, username types.Username)
...
+			message = "You have been invited to join the UCL ARC Services Portal by " + string(sponsor.Username)
+		}
+		messageInfo := graphmodels.NewInvitedUserMessageInfo()
+
+		messageInfo.SetCustomizedMessageBody(&message)
+		requestBody.SetInvitedUserMessageInfo(messageInfo)
+
+		_, err = c.client.Invitations().Post(ctx, requestBody, nil)
+		if err != nil {
+			if strings.Contains(err.Error(), "already exists") {
```
t-young31, line 173: how do we get here if we've checked if the user exists at the top of this function?

internal/handler/web/users.go
```diff
@@ -180,7 +181,7 @@ func (h *Handler) PostUsersInvite(ctx *gin.Context) {
 		return
 	}
 
-	if err := h.users.InviteUser(ctx, invite.Email, sponsor); err != nil {
+	if err := h.users.InviteUser(ctx, invite.Email, sponsor); err != nil && !strings.Contains(err.Error(), "does not exist") {
```
t-young31, line 184: why is a not-found error okay? 
(also should be `errors.Is` if it is needed)

deploy/dev/config.sample.yaml
```diff
@@ -13,6 +13,14 @@ entra:
   invite_redirect_url: <CHANGE_ME> # where you want to send the user when they accept the invite, eg. http://localhost:8000/profile
   primary_domain: <CHANGE_ME> # eg. arctretest.onmicrosoft.com
   invited_user_group_id: <CHANGE_ME> # found in the Azure portal, eg Invited Users group
+
+  # Mail app details, linked to staging on UCL tenant
+  # these are different for dev/staging to the details above but will be the same on prod
+  mail_tenant_id: <CHANGE_ME>
+  mail_client_id: <CHANGE_ME>
+  mail_client_secret: <CHANGE_ME>
+  mail_user_principal: <CHANGE_ME> # found in entra by searching for the mailbox
```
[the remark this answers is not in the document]
acholyn: if you didn't previously send an invite to that account and accept it, yes that's expected; the new mail tenant config is for handling custom mail via our mailbox, not using the built-in entra invite

internal/controller/entra/main.go
acholyn: tried with ccaetyo@... & tom.a.young@...
both triggered that path, but I don't see them in our TRE tenant? see img 
So I don't entirely understand why but it happens
image

It may be a peculiarity because in development we're using two different tenants?

deploy/dev/config.sample.yaml
```diff
@@ -13,6 +13,14 @@ entra:
   invite_redirect_url: <CHANGE_ME> # where you want to send the user when they accept the invite, eg. http://localhost:8000/profile
   primary_domain: <CHANGE_ME> # eg. arctretest.onmicrosoft.com
   invited_user_group_id: <CHANGE_ME> # found in the Azure portal, eg Invited Users group
+
+  # Mail app details, linked to staging on UCL tenant
```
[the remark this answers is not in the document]
t-young31: yeah – not sure. potentially just rm, as it's very UCL-specific and add it to the pile of 'what should this value be' in a new dev setup questions

internal/handler/web/users.go
acholyn: hmmm, I think possibly we could get away with it just being in `SendInvite`, I'll check

acholyn: nope, doesn't work just checking on the `SendInvite` side, I think because the error is returned at the top of that function with userData so never gets down to the `errors.Is` part

PR #274, opened 2025-09-05
making people page search-based (for admin + treops)

acholyn:
### Checklist

- slightly changed layout so search is more prominent as a page function
- split out length as a different err message in regex
image

- more clarity on regex err message
image

### DIFF
--- a/web/src/pages/people/index.tsx
+++ b/web/src/pages/people/index.tsx
@@ -141,6 +141,7 @@ export default function PeoplePage() {
       {!isAdmin && <Callout construction />}
       {canSearch &&
         searchTerm.length > 0 &&
+        !searchErrorMessage &&
         (!users || users.length === 0 ? (
           <Box>
             <div className={styles["no-users-found"]}>
@@ -155,8 +156,7 @@ export default function PeoplePage() {
             )}
           </Box>
         ) : (
-          !errorMessage &&
-          !searchErrorMessage && (
+          !errorMessage && (
             <>
               <h3 className={styles["results-heading"]}>Results for &ldquo;{searchTerm}&rdquo;</h3>
               <UserDataTable canEdit={isAdmin!} users={users} setUsers={setUsers} isLoading={isLoading} />


web/src/pages/people/index.tsx
```diff
@@ -44,9 +44,6 @@ export default function PeoplePage() {
       setIsLoading(false);
     }
   };
-  useEffect(() => {
```
acholyn, line 47: Some questions about how much of the functionality we want to remove:
- should I _remove_ all the fetching calls at this stage since they won't be needed until projects has been implemented, or just disable them?
- on the backend, should I remove the functions for getting all users/all approved researchers or is there motive to keep them for direct API calls?

jhughes982: On the frontend, if the initial fetching of people isn't needed anymore then i'd remove it. It can always be brought back from the git history if needed. We only need to run the fetch on a search now right?

For the backend, my feeling would be the same i.e. remove anything that isn't being used now as it can be reinstated later. But happy for you to wait for Tom's thoughts too if desired.

internal/handler/web/users.go
```diff
@@ -30,38 +30,28 @@ func (h *Handler) GetUsers(ctx *gin.Context, params openapi.GetUsersParams) {
...
 	} else if isTreOpsStaff {
-		h.getUsersTreOps(ctx, query)
+		h.getUsersTreOps(ctx, &query)
 
 	} else {
 		ctx.JSON(http.StatusInternalServerError, "Not implemented")
 	}
 }
 
 func (h *Handler) getUsersAdmin(ctx *gin.Context, query *string) {
```
t-young31, line 47: this and `getUsersTreOps` can now take `query` as a `string` (rather than `*query`) as it's required

jhughes982:
I wonder if the `Results for...` message could be a bit more prominent? Perhaps a bit of margin around it and a larger font size?

Screenshot 2025-09-09 at 15 43 31

jhughes982:
A nice to have feature could also be a 'clear search' button to reset the query. That might be preferable to the user having to delete the query term and pressing 'enter'. What do you think? Maybe not essential for this PR?

jhughes982:
Another thought, although this may not be needed and might need feedback from the users before building it: What do we think about more advanced search, for instance fetching all users with expired training?

jhughes982:
> > A nice to have feature could also be a 'clear search' button to reset the query. That might be preferable to the user having to delete the query term and pressing 'enter'. What do you think? Maybe not essential for this PR?
> 
> could do in another PR - would we want it as a sort of translucent x in a circle? a bit of text overhead? we'd need to sufficiently distinguish it from the submit button

Sounds good. Happy to see any designs you have in mind. For what it's worth, I'd probably just keep it simple and have some text that says 'clear search', make it blue or something so it looks like you can click it.

jhughes982:
> > Another thought, although this may not be needed and might need feedback from the users before building it: What do we think about more advanced search, for instance fetching all users with expired training?
> 
> I like that a lot but I'm not sure it would really be used 🤔 It maybe could be if they were going to go through quite a few users quickly and wanted to bring up the list and then crtl+f on the page? I think one to ask about but good to have in the back of our minds as potential

Yeah, really depends on how people intend to use the search.

PR #288, opened 2025-09-17
fix logout redirect url

### LINKED ISSUE #282: Log out still persists last user
After logging out and then logging back in, you're logged in as the last user. Would be good to know if this is something from our side or something to do with entra sessions. Not a massive problem atm but could be problematic on shared devices

t-young31:
## Resolves #282

- Redirect url needs to be query escaped and the redirect domain whitelisted. See: https://oauth2-proxy.github.io/oauth2-proxy/features/endpoints#sign-out

### Checklist


acholyn (approved):
Did this come up because of prod stuff or related to the previous user persisting after logout?

PR #315, opened 2025-10-01
check studyIDs when CRUDing asset contracts

### LINKED ISSUE #314: Always validate sutdyIds in the study service

t-young31:
### Checklist

### DIFF
--- a/internal/handler/web/studies.go
+++ b/internal/handler/web/studies.go
@@ -326,9 +326,6 @@ func (h *Handler) PostStudiesStudyIdAssetsAssetIdContractsUpload(ctx *gin.Contex
 		return
 	}
 
-	startDate := mustParseDate(contractMetadata.StartDate)
-	expiryDate := mustParseDate(contractMetadata.ExpiryDate)
-
 	user := middleware.GetUser(ctx)
 
 	// Open the uploaded file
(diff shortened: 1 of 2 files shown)


internal/handler/web/studies.go
```diff
@@ -215,19 +215,12 @@ func (h *Handler) PostStudiesStudyIdAssets(ctx *gin.Context, studyId string) {
 }
 
 func (h *Handler) GetStudiesStudyIdAssetsAssetId(ctx *gin.Context, studyId string, assetId string) {
-	studyUUID, err := parseUUIDOrSetError(ctx, studyId)
-	if err != nil {
-		return
-	}
-
-	assetUUID, err := parseUUIDOrSetError(ctx, assetId)
+	uuids, err := parseUUIDsOrSetError(ctx, studyId, assetId)
```
acholyn, line 218: I don't see this function anywhere else in this PR, should it be singular instead of plural?

jhughes982:
The changes here look good. I just have a few queries so that I can understand the purpose and intention here:
- Based on the title of the PR I assumed the changes would include checking if a study exists when creating an asset. But it looks like the check is made only when creating a contract or downloading a pdf contract. Is that the intention?
- Based on the preceding slack discussion to this PR, I anticipated this PR would include some RBAC changes. I can't currently see how these changes address the underlying issue discussed in the Slack thread.

I have probably just missed something, so any clarification would be great.

t-young31:
> Based on the title of the PR I assumed the changes would include checking if a study exists when creating an asset. But it looks like the check is made only when creating a contract or downloading a pdf contract. Is that the intention?

apologies – poor naming. it's in all places where the `studyID` wasn't being passed or checked  in the asset-contracts CRUD

> Based on the preceding slack discussion to this PR, I anticipated this PR would include some RBAC changes. I can't currently see how these changes address the underlying issue discussed in the Slack thread.

will respond on slack

PR #316, opened 2025-10-02
293/rework study flow

### LINKED ISSUE #293: Modify Study flow with final Submit & Review
**Currently** the Study's Status becomes "Pending" and it is available for review and feedback in the IG Admin's view when: 1) the Study form is submitted AND the 2) Study Agreement has been signed.

**Target** is that the Study's Status becomes "Pending" only when: 1) the Study form is submitted AND 2) the Study Agreement has been signed AND 3) 1 or more assets have been created AND 4) any required contracts have been uploaded AND 5) the user clicks Submit after all those steps. 

Any progress made with steps 1-4 are auto saved and the user can navigate away from the portal and pick back up their "Incomplete" Study in the /studies page.

Upon review the Study status will become "Approved" or "Changes Requested"

@acholyn have made up these statuses: can you sanity check and/or rename as you prefer? But does this largely sound ok?

acholyn:
### Checklist

drive by:
-clearing search
fixing dialog width changing
adding some max widths for some elems on study/asset pages


api.web.yaml
```diff
@@ -285,6 +285,24 @@ paths:
         default:
           description: Unexpected error
 
+  /studies/pending:
```
t-young31, line 288: I think `pending` should _probably_ be a query parameter on `/studies` as it's a filter

internal/handler/web/studies.go
```diff
@@ -491,3 +512,43 @@ func (h *Handler) PostStudiesAdminStudyIdReview(ctx *gin.Context, studyId string
...
+	review := openapi.StudyReview{
+		Status: openapi.Pending,
+	}
+
+	studyUUID, err := parseUUIDOrSetError(ctx, studyId)
+	if err != nil {
+		return
+	}
+
+	err = h.studies.UpdateStudyReview(studyUUID, review)
```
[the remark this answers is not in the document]
acholyn: keeping it in so admins know what to look at/for when re-reviewing

internal/handler/web/studies.go
```diff
@@ -491,3 +512,43 @@ func (h *Handler) PostStudiesAdminStudyIdReview(ctx *gin.Context, studyId string
...
+	studyUUID, err := parseUUIDOrSetError(ctx, studyId)
+	if err != nil {
+		return
+	}
+	studyData := openapi.StudyRequest{}
+	if err := bindJSONOrSetError(ctx, &studyData); err != nil {
+		return
+	}
+
+	log.Debug().Msgf("Updating study [%v]", studyUUID)
```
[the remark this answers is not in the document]
acholyn: do we want a debug message? a lot of these were from me trying to figure out why certain things weren't running

web/src/components/studies/ManageStudy.tsx
jhughes982, line 1: For a study that has a status of `incomplete`, if an admin tries to view that study they get a message: 

"To manage studies, please first set up your profile by completing the approved researcher process."

Is that intentional? Do we need to think more about how admins go through the approved researcher process?

jhughes982, line 1: For a study that has a status of `incomplete`, if an admin views that study and hovers over the status marker in the top right of the card, they see:

Screenshot 2025-10-07 at 16 06 17

The tooltip message does not reflect the current status since the study agreement was in fact agreed to in my case.

jhughes982, line 1: Based on the criteria listed in #293 a study should only be set to `pending` once we have fulfilled the following:

>Target is that the Study's Status becomes "Pending" only when: 1) the Study form is submitted AND 2) the Study Agreement has been signed AND 3) 1 or more assets have been created AND 4) any required contracts have been uploaded AND 5) the user clicks Submit after all those steps.

However, I was able to set the study to pending even though I had not yet uploaded a contract (I set the Asset as requiring one when I made it).

Would it be better to automate this somehow? So rather than requiring the user to manually set it for review, the status is automatically set once all the criteria have been fulfilled. This could be done with a simple function call perhaps.

jhughes982, line 1: As can be seen in the screenshot, it says `Study Setup Complete` even though I have not yet uploaded a contract for my Asset. It feels a bit misleading for anyone who is not used to the process. It feels like we need a way to highlight that the contract upload is incomplete.

Screenshot 2025-10-07 at 16 19 35

jhughes982 (commented):
Some initial comments as I am working through this. I'll be away from the computer for a bit so I thought I'd post this now to start the conversation. I'll continue with the review once I am back.

web/src/components/studies/ManageStudy.tsx
finleybacon: @acholyn i agree with your thinking on automation. it feels like if the ui is clear enough it shouldn't matter that we ask them to click a final big "Submit of approval" button. Maybe a "not yet submitted" tag/banner on the study tile could be helpful?

but notwithstanding that, @jhughes982 shouldn't have been able to set the study to pending even though he had not yet uploaded a contract which was required, right?

acholyn: yep, will add in bits for the contract req since that's also relevant to [another comment](https://github.com/ucl-arc-tre/portal/pull/316#discussion_r2411021624) made

finleybacon:
Ok - sorry for this - but how horrible is it to control f "Study Assets" and replace with "Information Assets" (at least in the UI).  And what are people's thought on that? @acholyn @jhughes982 @t-young31 ?

I think i prefer Information Assets since this is the language the ISMS uses for assets. Put another way, no other documentation i can think of calls them "Study assets". So this might prevent the overthinkers of the world from thinking Study Assets are a different thing.

web/src/components/studies/ManageStudy.tsx
jhughes982: Yeah, I don't feel strongly about automating the process. Making it clear in the UI should be sufficient, and we can implement additional safeguards such as: if the user clicks `Mark ready for review` but they haven't uploaded a contract, then a modal or error will tell them they need to upload a contract before submitting.

acholyn:
> Ok - sorry for this - but how horrible is it to control f "Study Assets" and replace with "Information Assets" (at least in the UI). And what are people's thought on that? @acholyn @jhughes982 @t-young31 ?
> 
> I think i prefer Information Assets since this is the language the ISMS uses for assets. Put another way, no other documentation i can think of calls them "Study assets". So this might prevent the overthinkers of the world from thinking Study Assets are a different thing.

fine by me; I did sort of wonder why we'd called them that but thought it also made sense given the hierarchy

will include as a drive by unless major objections?

api.web.yaml
```diff
@@ -338,6 +371,28 @@ paths:
...
+  /studies/{studyId}/pending:
+    post:
+      description: Update study status to pending
+      parameters:
+        - in: path
+          name: studyId
+          required: true
+          description: ID of the study
+          schema:
+            type: string
```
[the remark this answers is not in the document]
acholyn: I think I had done it as a more general function/route but realistically we're only looking for pending studies and all studies (afaia) so chose not to add that complexity. Happy to change it or can leave it till we have use of it. @t-young31 @finleybacon , thoughts?

web/src/components/studies/ManageStudy.tsx
acholyn: https://github.com/ucl-arc-tre/portal/pull/316/commits/17f37801d5b62f6381712dc8fc3a51a38f69352f

api.web.yaml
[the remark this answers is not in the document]
t-young31: `PATCH` probably makes sense, maybe as `/studies/{studyId}/status/pending`. study owners should only be able to set it as pending (i.e. not `approved` or `rejected`), so I don't think a generic endpoint is required

internal/handler/web/studies.go
```diff
@@ -121,12 +121,24 @@ func (h *Handler) GetStudies(ctx *gin.Context) {
...
+			}
+
+		} else {
+			// Admins can see all studies normally
+			studies, err = h.studies.AllStudies()
+			if err != nil {
+				setError(ctx, err, "Failed to get studies")
+				return
+			}
 		}
```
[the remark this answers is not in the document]
t-young31: I think a query parameter here makes sense as it's a filter of the response. agree that the function is getting a little complex – wouldn't be adverse to splitting it into `getStudiesAdmin(ctx, params)` and `getStudiesStudyOwner(ctx)`, depending on the role

web/src/pages/studies/index.tsx
jhughes982, line 1: As far as I can tell, the `fetchStudies` function has no advantage of being called in this component anymore, since the data isn't used until it is passed down to the `Studies` component. I'd consider moving the fetch to the component where it is actually used. Otherwise you risk making unnecessary network requests where the data isn't utilised.

internal/service/studies/main.go
```diff
@@ -242,8 +247,23 @@ func (s *Service) UpdateStudy(id uuid.UUID, studyData openapi.StudyRequest) erro
...
+	}
+
+	for _, studyAdminUser := range studyAdmins {
+		studyAdmin := types.StudyAdmin{
+			StudyID: study.ID,
+			UserID:  studyAdminUser.ID,
+		}
+		if err := s.db.Create(&studyAdmin).Error; err != nil {
+			return types.NewErrFromGorm(err, "failed to create study admin")
+		}
```
t-young31: should be idempotent

acholyn: have created https://github.com/ucl-arc-tre/portal/issues/335 since it's a bigger problem and should be handled in a separate PR

jhughes982:
> > . I logged in as a staff member and became an approved researcher, but I seem to have lost the studies navlink in the navbar and navigating manually to /studies gives me the message You haven't been added to any studies yet.
> 
> Will see if the same thing happens for me, but that's not expected behaviour

This is fixed now, unknown exactly what caused it but a reset of my environment fixed it.

jhughes982:
Following on from my above comment about administrators. Is there any reason why we don't allow removing administrators once the study has been submitted? What if an administrator leaves and the study needs to assign a different administrator?

If we really do want to prevent removing administrators, then some explanation of this in the UI would be helpful.

acholyn:
>Is there any reason why we don't allow removing administrators once the study has been submitted? What if an administrator leaves and the study needs to assign a different administrator?

We've temporarily disabled it until we do #335 since we haven't got the backend on the role (IAA) and haven't added the ability to delete people/remove roles yet and seemed like it would be better to tackle in another PR

acholyn:
>The check for duplicate study names is not case sensitive, so I can submit two studies with the names study 1 and Study 1. Is that what we want (we may have already discussed this, but I can't quite remember)?

Ah, we don't want to allow this. Will tackle in another PR, have created #352

internal/service/studies/main.go
```diff
@@ -216,7 +216,7 @@ func (s *Service) createStudy(owner types.User, studyData openapi.StudyRequest,
 			StudyID: study.ID,
 			UserID:  studyAdminUser.ID,
 		}
-		if err := tx.Create(&studyAdmin).Error; err != nil {
+		if err := tx.FirstOrCreate(&studyAdmin).Error; err != nil {
```
jhughes982: Is this solution working on your end? I'm still getting the same bug unfortunately.

web/src/components/studies/StudyForm.tsx
```diff
@@ -51,34 +51,34 @@ const convertStudyFormDataToApiRequest = (data: StudyFormData) => {
...
+    involves_external_users: data.involvesExternalUsers ? data.involvesExternalUsers : false,
+    involves_participant_consent: data.involvesParticipantConsent ? data.involvesParticipantConsent : false,
     involves_indirect_data_collection: data.involvesIndirectDataCollection
       ? data.involvesIndirectDataCollection
-      : undefined,
+      : false,
     involves_data_processing_outside_eea: data.involvesDataProcessingOutsideEea
       ? data.involvesDataProcessingOutsideEea
-      : undefined,
+      : false,
```
jhughes982: There are still issues with this approach if the fields are no longer optional:
1. This needs to be reflected in `api.web.yaml`
2. The `go` type for a study may need to be updated as the model currently suggests some of the fields can be `nil` (i.e. optional)
3. The above solution does not address the follow-on questions that require text input, e.g. the `iras_id`
4. Updating the `data_protection_number` field leaves a permanent record in the DB, even if I uncheck `is_data_protection_office_registered ` later on.

Probably a bit more thinking is needed around how this works. Maybe @t-young31 can advise further on representing these values as `go` types?

jhughes982:
If I submit my study for review, the blue button still says 'Mark Ready for Review'. I can also spam this button to trigger as many network requests as I like. I think the UX could be improved here by either disabling the button entirely (to prevent the network requests going out), or at least to make the scenario more clear to the user. For instance, the button could be greyed out and the text could say e.g. 'Study submitted for review'. Or the button could be removed with the status being presented more obviously to the user, i.e. `Your study is currently under review. If you need to make changes click the edit button and then resubmit the study`. 

I think it is worth being really clear to the user about their study status to avoid confusion.

web/src/components/studies/StudyForm.tsx
jhughes982: I spotted something else that may be related to this conversation:
- I filled out a study with the `Data Protection Registration Number` filled out,
- I submitted the study and marked it for review,
- I then went to edit the study and changed the `Data Controller` text,
- As I went through the next study pages the `Data Protection Registration Number` was now blank and it wouldn't let me submit until I had filled out the field again.

Weirdly, this only happened when I updated the `Data Controller` input text. Changing other parts of the form didn't produce the above behaviour.

acholyn:
> I would make it so that a modal appears when the admin clicks `approve` and the modal should say something like `Are you sure...`.

It's expected that most studies will be approved, I'm not sure it's a big enough deal to make them go through another click when we don't expect them to be flippant about it, but @finleybacon @t-young31 what do you reckon?

finleybacon:
@acholyn @jhughes982 thanks for waiting, so IG Admins do want to see certain asset properties. Looks like an "approve with feedback" option might be nice as well.

Matthew's comment I think is more about being able to search/query all studies and assets for reporting purposes. 

Finley
> We are proposing a single review of the Study (so study details + contracts + assets). So the question is about what you'll need to see to do that review:
> 1. Do you need to see a list of all the assets that have been created in the Study?
> 2. If so, what properties of each asset do you need to see?
> 3. Do you need to read the third party contracts if the user has uploaded it? (I'm assuming no, you don't need to read it/them, but is there some metadata about the contract(s) that you want to know?)
 

Victor
> For the asset register, we consider classification, impact mitigation, current location and legal basis. These factors are used to advise researchers on risk mitigation and data practises.

Victor
> We review the contents of contracts on a very rare occasion but do check the expiry or review date and the UCL signatory. I would like to request that the new system send reminders to both the researchers and create a MyServices note for contracts that are due for expiry. This would help us avoid holding data beyond the stipulated contract period.

Matthew Lilliman
> Can I also add that searching these records is important, although it's a lesser need to those of running the IG service  In the last few weeks I've had a handful of NHSE emails via UCL's SIRO asking if data have been/will be destroyed.  From a PI name, study name and contract name I need to track down the study in the IG portal to start talking to the right team, it should be simple, but the PI is not always the IAO/IAA, the study has a different name and the name on the contract in the IG portal may not match.  Eventually I have been tracking the UCL team down to get things resolved, but having everything currently searchable in the IG portal has proved v-useful.

Preeti
> Hello Finley Bacon, a single review of the study sounds like a great idea - it would save a considerable amount of time from our side.

web/src/components/studies/StudyForm.tsx
acholyn: the data controller is a required field and if you changed it, that affects the data protection registration number since the latter is based on the former
I think that behaviour is largely expected, unless I'm missing something

jhughes982: I see, I wasn't aware of the coupling between the two fields. If they are closely linked I wonder if it would be better to present them on the same form step (for a different PR though)?

acholyn:
> What do you think?

will tackle in a separate PR as approve, approve + feedback, reject + feedback

internal/service/studies/main.go
```diff
@@ -214,6 +245,58 @@ func (s *Service) UpdateStudyReview(id uuid.UUID, review openapi.StudyReview) er
...
+	}
+	study := studies[0]
+	setStudyFromStudyData(&study, studyData)
+
+	studyAdmins, err := s.createStudyAdmins(studyData)
+	if err != nil {
+		return err
+	}
+
+	for _, studyAdminUser := range studyAdmins {
```
t-young31: copy pasta sadness

internal/controller/entra/custom_mail.go
```diff
@@ -105,9 +105,9 @@ func (c *Controller) SendCustomInviteNotification(ctx context.Context, email str
 	return c.createCustomEmail(ctx, "Notification: You have been invited to the UCL ARC Services Portal", []string{email}, content)
 }
 
-func (c *Controller) SendCustomStudyReviewNotification(ctx context.Context, email string, review openapi.StudyReview) error {
+func (c *Controller) SendCustomStudyReviewNotification(ctx context.Context, emails string, review openapi.StudyReview) error {
 	// email to notify IAO + IAA when study has been reviewed
-	// todo: set up usage of this
+	// not 100% sure if this works due to which tenant we match admins against + arctretest.onmicrosoft inboxes??
```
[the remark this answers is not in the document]
acholyn: need to test it on staging basically
tre tenant uses "@arctest.onmicrosoft.com" as the domain, so if you add an admin they have to be staff on that but our usernames (portal-ahl-admin) don't actually have email boxes that we can access (afaik) so I can't test locally if I receive an email because there is no place for that email to go but I can't send to a real email because that doesn't have staff status

web/src/components/studies/StudyForm.tsx
```diff
@@ -2,74 +2,150 @@ import { useState, useEffect } from "react";
...
+  const studyData: StudyRequest = {
+    title: data.title,
+    description: data.description ? data.description : undefined,
+    data_controller_organisation: data.dataControllerOrganisation.toLowerCase(),
+    additional_study_admin_usernames: data.additionalStudyAdminUsernames
+      .map((admin) => admin.value.trim())
+      .map((username) => `${username}${domainName}`),
+    involves_ucl_sponsorship: data.involvesUclSponsorship !== undefined ? data.involvesUclSponsorship : undefined,
+    involves_cag: data.involvesCag !== undefined ? data.involvesCag : undefined,
+    cag_reference: data.cagReference ? data.cagReference.toString() : undefined,
```
jhughes982: If I submit a study with `involves_cag: true` a `cag_reference`, but then later edit my study to `involves_cag: false` the original `cag_reference` will remain. 

I'm not sure what the desired behaviour is here, but it's worth flagging now in case we want to remove any previously entered references.

I think this would also apply to the other conditional inputs too.

web/src/pages/studies/manage.tsx
```diff
@@ -25,13 +25,13 @@ export default function ManageStudyPage() {
...
 
-  const fetchStudy = async (id: string) => {
+  const fetchStudy = async (id?: string) => {
     setStudyIsLoading(true);
     setStudyError(null);
 
     try {
       const response = await getStudiesByStudyId({
-        path: { studyId: id },
+        path: { studyId: id! },
```
jhughes982: This is one of those times where the use of `!` is not appropriate. As discussed previously, in TypeScript you really want to avoid using this unless you really have to as it bypasses the safety checks that TypeScript is designed to give you.

In this case we get an interesting effect. I can modify the URL to be this: `http://localhost:8000/studies/manage?studyId=` (so I'm omitting the study id string). This will cause the following code to run...

```
  useEffect(() => {
    if (studyId == undefined || !isApprovedResearcher) return;
    fetchStudy(studyId as string);
  }, [studyId, isApprovedResearcher]);
```

...because `undefined` will not catch the empty string. What happens in this scenario is that the frontend will call the api endpoint to fetch all studies at `/web/api/v0/studies` which is definitely not what we want here. 

So the guard clauses need tweaking here.

web/src/pages/assets/manage.tsx
```diff
@@ -23,6 +23,8 @@ export default function ManageAssetPage() {
   const [error, setError] = useState<string | null>(null);
 
   const isApprovedResearcher = userData?.roles.includes("approved-researcher");
+  const isStudyOwner =
+    (userData?.roles.includes("information-asset-owner") && study?.owner_username === userData.username) || false;
```
jhughes982: I think the `|| false` is redundant here?

web/cypress/e2e/pages/study-management.cy.ts
```diff
@@ -153,8 +153,51 @@ describe("Study Assets Management", () => {
...
+    cy.contains("Edit Study").click();
+
+    // edit the description and submit
+    cy.get("textarea#description").type("My New Test Study Description");
+    cy.get("[data-cy='next']").click();
+    cy.get("[data-cy='next']").click();
+
+    cy.get("button[type='submit']").click();
+
+    cy.wait("@updateStudy");
```
jhughes982: I think as soon as this PR is merged we should prioritise https://github.com/orgs/ucl-arc-tre/projects/1/views/15?pane=issue&itemId=136189709&issue=ucl-arc-tre%7Cportal%7C354 to allow more robust testing of the functionality. Then we should come back and replace some of these tests with true e2e versions.

jhughes982 (changes requested):
Still a bit more work needed before this should be merged in my opinion. Some comments can be moved into issues and tackled in follow-up PR's but a few of them I think should definitely be done in this PR.

web/src/components/studies/StudyForm.tsx
```diff
@@ -2,74 +2,150 @@ import { useState, useEffect } from "react";
...
+  editingStudy?: Study | null;
 };
 
-// what is this?
-// A brief comment for context might be good here
+const convertStudyFormDataToApiRequest = (data: StudyFormData) => {
+  console.log("cagref value", data.cagReference);
+  const studyData: StudyRequest = {
+    title: data.title,
+    description: data.description ? data.description : undefined,
```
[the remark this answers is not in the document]
acholyn: Description isn't required so it's fine to have it as an empty string

web/src/components/studies/StudyForm.tsx
```diff
@@ -142,20 +218,113 @@ export default function CreateStudyForm(CreateStudyProps: CreateStudyProps) {
...
+      });
+    }
+  }, [editingStudy, username, reset]);
+
+  const handleStudySubmit = async (data: StudyFormData, studyId?: string) => {
+    setIsSubmitting(true);
+    setSubmitError(null);
+
+    try {
+      // Convert form data to API format
```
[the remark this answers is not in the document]
acholyn: https://github.com/ucl-arc-tre/portal/pull/316/commits/6a0e7330ad8b64b15634445017bc003aaa078338

web/src/pages/studies/manage.tsx
acholyn: https://github.com/ucl-arc-tre/portal/pull/316/commits/e289a181a7ec35f894d8e8dc9e0d08d91bcf27d9 ?

web/src/components/studies/StudyForm.tsx
acholyn: https://github.com/ucl-arc-tre/portal/pull/316/commits/e5263bb73dea7d8b042a37aec9a61fcd0091e394

web/src/components/studies/StudyForm.tsx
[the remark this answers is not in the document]
jhughes982: > it's fine to have it as an empty string

I personally disagree. In the case of the user entering a completely empty string (which would highly likely not be intentional by the user) I don't think we should be storing blank data in the database. If the field is completely empty I'd opt for trimming the data and storing this as `null`. This way it keeps the database entries consistent so we don't end up with a mixture of empty strings and/or null values.

I'll leave the final decision to you though, and happy for it to be in a follow-up PR if you decide to go ahead with it.

web/src/components/studies/Studies.tsx
```diff
@@ -1,117 +1,118 @@
...
+            className={`${styles.tab} ${styles["all-studies-tab"]} ${tab === "all" ? styles.active : ""}`}
+          >
+            All Studies
+          </Button>
+        </div>
+
+        {studiesLoading && <Loading message="Loading studies..." />}
+
+        <StudyCardsList studies={tab === "pending" ? pendingStudies : studies} isAdmin={true} />
+      </>
```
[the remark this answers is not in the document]
jhughes982: The below might be personal preference, but something that I think is easy to do and improves UX:

This does now display a `no studies found` message in the case of there literally being no studies in the database at all. But it doesn't display the message separately for Pending studies and All Studies. That means that if a study has been created but not yet flagged for review, the Pending studies list just shows a blank space and there is no further feedback e.g. `No pending studies found`. I would show a custom message separately for both Pending and All studies.

In the interest of not prolonging this PR further, perhaps this can go in a follow-up issue ticket?

web/src/pages/studies/manage.tsx
jhughes982: Looking at this again, I'm not sure I actually understand its intention. 

```
        const response = await getStudiesByStudyId({
          path: { studyId: id },
        });
```

It looks like the above is fetching a specific study by its `id`, so why does `const fetchStudy = async (id?: string) => {` take in an optional `id`? Surely the `id` would always be necessary for fetching the specific study right?

web/src/components/studies/StudyForm.tsx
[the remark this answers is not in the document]
acholyn: ahh, I see the distinction
yes, may have to be a follow up as it seems just setting to null isn't working 🤔 

opened #385

web/src/components/studies/StudyForm.tsx
acholyn: > If I enter a reference number for the optional fields, but then set the parent field to no then the reference number remains in the database.

I think this is related to #385 

> It is impossible to set a form field back to unsure if any other answer was given previously, is this intentional?

no, but I think it's a side effect of not being able to update a value to be undefined

PR #366, opened 2025-11-07
Project routes

### LINKED ISSUE #10: Project creation
## Description

**As an**  information asset owner or administrator
**I want to** create a project
**So that** I can use a UCL TRE for my research

## Acceptance criteria

- [ ] Users can only create projects for approved studies - done through manage study page
  - [ ] optional: create project option on study card for quick create
- [ ] users will have options to choose assets for the project, which would then disable some environment options (explanation on hover) AND can choose env first without assets, which would then disable some asset options
  - [ ] Project cannot be created with assets classified with a tier higher than the environment
- [ ] TRE Ops staff can update project creation status for TRE project
  - [ ] statuses: draft (saved on create form) | submitted (when submitted on create/manage) | deployed (only updatable by TRE ops staff)
- [ ] Users can save a draft of their project record and return to it later (similar to studies)
- [ ] In creating a Project record the user must specify Approved Researchers (similar to studies) ~~or invite certain users to become an Approved Researcher (#2 and/or #3 )~~
  - [ ] a project can't be submitted unless all users are valid. it can be saved if on create, otherwise, on manage only submit
  - [ ] if an inputted user is invalid, error message to tell IAO that they have been requested via email to do the things to become valid. this also happens on manage if they are invalid
- [ ] Projects will reference Information Asset records (#7)
- [ ] Project struct: see https://github.com/ucl-arc-tre/portal/issues/10#issuecomment-3426812380 and https://github.com/ucl-arc-tre/portal/issues/10#issuecomment-3431207964
- [x] TREProject inherits from Project
- [ ] TREProjectRoleBinding model to link projects to users (like studyadmins) with string roles
- [ ] TRE ops staff would have access to all TRE projects, and would be able to view /api/projects/tre/<uuid>
  - [ ] GET TreProject preload Projects
- [ ] Project members should be allowed to view assets & contracts but not edit or download

## Notes

- Project can only be created in an environment {TRE, DSH, Condenser} which is suitable for the study data
- Tick boxes for functionality (e.g. RedCap) that narrows choice of environment - TBC
- TRE Ops process to be decided: Certain permissions groups need to be selected per user?

## Dependencies

- [x] #2
- [x] #297 
- [x] #3 
- [x] #5 
- [x] #7

jhughes982:
## Resolves #10 

- This PR adds in a few basic project routes for getting and posting a project (restricted to TRE for now),
- The associated Go handlers have also been added (with dummy logic for now),
- The frontend now talks to the above handlers,
- The DB model has also been initialised,
- Very basic RBAC has also been added for approved researcher. This will be expanded in the next stages when adding dynamic roles for project owners etc.

### Checklist

### DIFF
--- a/internal/graceful/db.go
+++ b/internal/graceful/db.go
@@ -30,8 +30,6 @@ func InitDB() {
 		&types.Contract{},
 		&types.UserSponsorship{},
 		&types.Environment{},
-		&types.Project{},
-		&types.ProjectTRE{},
 	}
 	db := NewDB()
 	db.Exec(`CREATE EXTENSION IF NOT EXISTS "uuid-ossp";`)


t-young31 (approved):
🚢

PR #367, opened 2025-11-07
Project rbac

jhughes982:
## Works towards #10 

- This PR adds a bit more detail to the project RBAC by adding dynamic access (which will eventually be called at the point of project creation/requesting projects).
- I think this might need further conversations to get this right as I am not sure that the RBAC I have added is entirely what everyone had in mind at the moment. Happy to adjust as needed.
- I have also started an abstraction of the dynamic RBAC logic, feedback would be welcome here as to whether the abstraction is a useful one, or even if further abstraction would be better (e.g. shall we abstract all of the dynamic RBAC logic so that it is agnostic of studies/projects?).

### Checklist


internal/rbac/main.go
```diff
@@ -90,6 +71,30 @@ func StudyIDsWithRole(user types.User, studyRoleName StudyRoleName) ([]uuid.UUID
...
 }
 
+// Add a project TRE owner role for a user
+func AddProjectTreOwnerRole(user types.User, projectId uuid.UUID) (bool, error) {
+	roleName := makeProjectOwnerRole(projectId).RoleName()
+	return addOwnerRoleForResource(user, roleName, []string{"/projects/tre"}, projectId)
+}
+
+// Project IDs where a user has a role
+func ProjectIDsWithRole(user types.User, projectRoleName ProjectRoleName) ([]uuid.UUID, error) {
```
jhughes982: This can potentially be abstracted to combine the logic with `StudyIDsWithRole`. But I wanted to hear peoples thoughts first as I'm not sure if the project environments complicates this or not, and therefore whether separating the logic is overall more simple to follow?

PR #376, opened 2025-11-12
Project form starter

jhughes982:
## Works towards #10 

- This PR adds the first stages of the project form creation. This is entirely frontend for now so no additional backend logic has been added at this stage.
- I've kept it light for now as I think this already sparks some discussion over the direction the form will go in, so we can discuss now before it gets more complex.
- I have started building this with future use cases/environments in mind so that we can keep this flexible from the start as opposed to restricting it specifically to the TRE use case.
- @finleybacon after thinking about this, I think your idea of presenting the environment before the assets is better. This is because certain types of projects can be submitted without having an associated asset. So I think the assets can come after environment selection as an optional field and those assets will be restricted based on the environment they chose (with appropriate UI feedback so they know exactly whats going on). Open to more ideas and discussions here though.
- As far as I am aware, having at least 1 study is still a pre requisite to making a project though, so I added this as a requirement.
- I also now allow the navbar to show the `projects` nav item even before the user has made an asset. This is because [previous discussions](https://github.com/ucl-arc-tre/portal/issues/10#issuecomment-3503280900) have suggested some types of projects can be made before becoming an Information Asset owner. The only requirement is that the user agrees to the study agreement beforehand.
- Next up will be adding more project form features and starting to hook things up to the backend.
- Tests will be added once more of the functionality has been signed off.

### Checklist

### DIFF
--- a/web/src/components/projects/CreateProjectForm.tsx
+++ b/web/src/components/projects/CreateProjectForm.tsx
@@ -14,12 +14,12 @@ type ProjectFormData = {
 };
 
 type Props = {
-  studies: Study[];
+  approvedStudies: Study[];
   handleProjectCreated: () => void;
   handleCancelCreate: () => void;
 };
 
-export default function CreateProjectForm({ studies, handleProjectCreated, handleCancelCreate }: Props) {
+export default function CreateProjectForm({ approvedStudies, handleProjectCreated, handleCancelCreate }: Props) {
   const [isSubmitting, setIsSubmitting] = useState(false);
   const [error, setError] = useState<string | null>(null);
 
@@ -94,13 +94,11 @@ export default function CreateProjectForm({ studies, handleProjectCreated, handl
               disabled={isSubmitting}
             >
               <option value="">Select a study...</option>
-              {studies
-                .filter((study) => study.approval_status === "Approved")
-                .map((study) => (
-                  <option key={study.id} value={study.id}>
-                    {study.title}
-                  </option>
-                ))}
+              {approvedStudies.map((approvedStudy) => (
+                <option key={approvedStudy.id} value={approvedStudy.id}>
+                  {approvedStudy.title}
+                </option>
+              ))}
             </select>
             {errors.studyId && (
               <Alert type="error">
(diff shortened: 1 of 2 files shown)


web/src/components/projects/CreateProjectForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+            study_id: data.studyId,
+          };
+          response = await postProjectsTre({ body: requestBody });
+          break;
+        }
+        case "dsh":
+          throw new Error("DSH projects are not yet supported");
+        default:
+          throw new Error(`Unknown environment: ${data.environment}`);
+      }
```
jhughes982: Depending on how many environments we end up with, we might want a neater solution to this. Maybe some sort of lookup that contains more project attributes for each environment.

web/src/components/projects/CreateProjectForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+            <select
+              id="studyId"
+              className={styles.select}
+              {...register("studyId", {
+                required: "Please select a study",
+              })}
+              disabled={isSubmitting}
+            >
+              <option value="">Select a study...</option>
+              {studies.map((study) => (
```
acholyn: should only approved studies in the list

acholyn (changes requested):
A couple of tweaks to do but looks good so far 🎉

web/src/components/projects/CreateProjectForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+          <div className={styles["form-field"]}>
+            <label htmlFor="name">Project Name *</label>
+            <input
+              id="name"
+              type="text"
+              placeholder="e.g., my-project"
+              {...register("name", {
+                required: "Project name is required",
+                pattern: {
+                  value: /^[a-z0-9][a-z0-9-]*[a-z0-9]$/,
```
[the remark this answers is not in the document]
jhughes982: [moved project name field](https://github.com/ucl-arc-tre/portal/pull/376/commits/49dca079602bfb94016735712f2849734396238d)

PR #382, opened 2025-11-17
added asset selection to projects form

### LINKED ISSUE #10: Project creation

jhughes982:
## Resolves #10 

- This PR adds optional asset selection to the projects form,
- Assets are fetched dynamically based on the study that is chosen,
- I use the `environments` table in the database as the source of truth here. This does mean I needed to add an environments endpoint and also modify the RBAC to allow approved researchers to access the endpoint. So let me know if you disagree with this approach or whether this needs tweaking further.
- I've designed this so that the user gets feedback on what is going on with their selections, so if they pick an environment that is incompatible with an asset the UI will tell them.
- Tests will come up next if this covers the full projects flow. Then I'll integrate this into the backend with full project DB submission.

## Note on selecting a study
- Since #316 is not quite ready, you will have to manually edit the DB entry for a submitted study to set its status to 'Approved' in order to create a project.

### Checklist


internal/service/environments/main.go
```diff
@@ -0,0 +1,17 @@
+package environments
+
+import (
+	"github.com/ucl-arc-tre/portal/internal/graceful"
+	"github.com/ucl-arc-tre/portal/internal/types"
+)
+
+func GetAll() ([]types.Environment, error) {
+	db := graceful.NewDB()
```
t-young31: this does the init call on every function call. I think we should use the standard pattern here of: the service is a struct https://github.com/ucl-arc-tre/portal/blob/068fe2c3338bff7ee087acb79046bf1ce7d0790f/internal/service/agreements/main.go#L16-L18
then the handler gets a new service https://github.com/ucl-arc-tre/portal/blob/068fe2c3338bff7ee087acb79046bf1ce7d0790f/internal/handler/web/main.go#L16-L17

web/src/components/projects/CreateProjectForm.tsx
```diff
@@ -109,29 +194,81 @@ export default function CreateProjectForm({ approvedStudies, handleProjectCreate
...
+              <option value="">
+                {isLoadingEnvironments
+                  ? "Loading environments..."
+                  : environmentsError
+                    ? "Error loading environments"
+                    : "Select an environment..."}
+              </option>
+              {environments.map((environment) => (
+                <option key={environment.id} value={environment.id}>
+                  {environment.name} (Tier {environment.tier})
```
t-young31: would be nice here to have the links to some documentation as some helper text once selected

PR #388, opened 2025-11-19
clearing values in db

### LINKED ISSUE #384: deleting data from optional fields
whilst we can set certain values to false, the IDs are not removed from the db

> [e5263bb](https://github.com/ucl-arc-tre/portal/commit/e5263bb73dea7d8b042a37aec9a61fcd0091e394)

This doesn't fix the issue unfortunately. I tested out various combinations of setting form fields to `true`, `false`, and `unknown` but the form doesn't behave as expected in a few different ways, e.g.:
- It is impossible to set a form field back to `unsure` if any other answer was given previously, is this intentional?
- If I enter a reference number for the optional fields, but then set the parent field to `no` then the reference number remains in the database.

As said in a previous comment, these kinds of edge cases can be caught more easily with a more robust testing methodology, which is why I think #354 should be prioritised.

_Originally posted by @jhughes982 in https://github.com/ucl-arc-tre/portal/pull/316#discussion_r2528236891_

### LINKED ISSUE #385: removing values during update
Some study fields (the optional IDs) and the description, can be set to null but the values aren't removed in the db

> it's fine to have it as an empty string

I personally disagree. In the case of the user entering a completely empty string (which would highly likely not be intentional by the user) I don't think we should be storing blank data in the database. If the field is completely empty I'd opt for trimming the data and storing this as `null`. This way it keeps the database entries consistent so we don't end up with a mixture of empty strings and/or null values.

I'll leave the final decision to you though, and happy for it to be in a follow-up PR if you decide to go ahead with it.

_Originally posted by @jhughes982 in https://github.com/ucl-arc-tre/portal/pull/316#discussion_r2526846430_

acholyn:
### Checklist

When conditional option fields (involves IRAS, involves NHSE, involves CAG, DPO registered) are set to false, the associated ID is cleared in the db (set to nil) [#384]
When description is deleted, it's set to nil [#385]

### DIFF
--- a/internal/service/studies/main.go
+++ b/internal/service/studies/main.go
@@ -163,7 +163,6 @@ func (s *Service) StudiesById(ids ...uuid.UUID) ([]types.Study, error) {
 
 // given a study and data from a request, line up the values
 func setStudyFromStudyData(study *types.Study, studyData openapi.StudyRequest) {
-	falseVal := false
 	if studyData.Title != "" {
 		study.Title = studyData.Title
 	}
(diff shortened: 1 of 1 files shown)


internal/service/studies/main.go
```diff
@@ -163,25 +163,42 @@ func (s *Service) StudiesById(ids ...uuid.UUID) ([]types.Study, error) {
...
 	if studyData.Title != "" {
 		study.Title = studyData.Title
 	}
-	study.DataControllerOrganisation = studyData.DataControllerOrganisation
 	study.Description = studyData.Description
+	study.DataControllerOrganisation = studyData.DataControllerOrganisation
 	study.InvolvesUclSponsorship = studyData.InvolvesUclSponsorship
 	study.InvolvesCag = studyData.InvolvesCag
-	study.CagReference = studyData.CagReference
+	if studyData.InvolvesCag == &falseVal {
```
t-young31: you can't compare pointers

t-young31 (approved):
🚢

Issue #372, opened 2025-11-11
NHS E reference number


acholyn:
What format does this number take so we can better validate; will it ever lead or trail with 0s?
---
For `nhs_england_reference` I can enter e.g. `00123`. In this case, the database does store `00123` but the frontend will display `123`. I'm not sure if there is a standard format used for this reference number, or even if leading zeros are allowed/possible. But we should make sure we are storing and displaying things correctly to avoid future confusion.

Also open to this being in a follow-up PR since presumably this issue existed before?

_Originally posted by @jhughes982 in https://github.com/ucl-arc-tre/portal/pull/316#discussion_r2510500939_

acholyn:
seems it's `DARS-NIC-xxxxxx-xxxxx-xx`
https://www.infosec.ox.ac.uk/apply-for-data-from-nhs-digital#:~:text=DARS%2DNIC%2Dxxxxxx%2Dxxxxx%2Dxx

acholyn:
CAG is xx/CAG/xxxx

IRAS is 7 digits

PR #392, opened 2025-11-24
DPN validation

### LINKED ISSUE #373: Data Protection # format
need to clarify the format
---
The help text for this says the format of the number is `Z6364106/2022/01/123` but we store it in the database as `Z6364106/2022-01/123`. I'm not sure how big of a deal this is, but we should at least show the true format in the help text.

Also, leading zeros for the final part are not displayed in the UI.

_Originally posted by @jhughes982 in https://github.com/ucl-arc-tre/portal/pull/316#discussion_r2510533480_

### LINKED ISSUE #380: Data protection registration validation
- In the create study form, the help text for the data protection registration number says the last segment should be a 2-3 digit number. But I am able to submit a single digit (even 0) and the form will accept this and store it in the database.
- We should match the help text to the actual requirements and make sure validation captures it correctly.

Image

tagging @acholyn for visibility

Edit:
- I can also enter in numbers like `0000123` into the form. In this case the database will actually store `0000123` (which should not be allowed based on the help text guidance) but when the data is displayed on the frontend form the number becomes `123` again.

acholyn:
### Checklist

addresses issues with DPN field

follow ups:
#372


jhughes982:
Is this going to require a migration?

t-young31:
> Is this going to require a migration?

prod has no studies with 0-prefixed DPNs, so think okay

acholyn:
> > > Is this going to require a migration?
> > 
> > 
> > prod has no studies with 0-prefixed DPNs, so think okay
> 
> What about the format, I think any previous entries will be stored as e.g. `Z123456/2025-11/123` whereas this PR changes it to e.g. `Z123456/2025/11/123`. Are we going to need to account for this?

hm..yes...could do a check for if the number has a - in it and adjust the slicing? when saved again the format should be updated

t-young31:
> hm..yes...could do a check for if the number has a - in it and adjust the slicing? when saved again the format should be updated

I'd prefer database migration and enforce strict regex rather than silently changing formats – think the latter could get messy quickly

internal/service/studies/main.go
```diff
@@ -21,7 +21,8 @@ import (
 )
 
 var (
-	titlePattern = regexp.MustCompile(`^\w[\w\s\-]{2,48}\w$`)
+	titlePattern                = regexp.MustCompile(`^\w[\w\s\-]{2,48}\w$`)
+	dataProtectionNumberPattern = regexp.MustCompile(`^\w+\/\d{4}\/\d{2}\/(?:(?:0[1-9])|(?:[1-9]\d{1,2}))$`)
```
t-young31: 🙃

t-young31:
@jhughes982 / @acholyn would you mind taking another look at this with [502a168](https://github.com/ucl-arc-tre/portal/pull/392/commits/502a16897f9be8fb0705d45ce1db652b11bc0195) added?

t-young31:
> Is this the behaviour we want?

from @finleybacon's  table of values that already exists this is desired

jhughes982:
> > Is this the behaviour we want?
> 
> from @finleybacon's table of values that already exists this is desired

Happy to go with that then. I do find it a bit jarring that the database saves my value as `01` but I am prevented as a user from manually entering `01`. But perhaps this is something users will never notice (since they won't see what's saved in the database)?

t-young31:
> the user the number cannot start with 0

trouble is it can. numbers <10 need to start with a 0 but >=10 can't. personally I find this formatting mad, which I'd rather not expose the madness to users too much. other options welcome

edit: alternatively we make the frontend text and let users deal with the mad format 🤷🏼

jhughes982 (approved):
Looks good. Perhaps @acholyn has some final thoughts on the format complexity and how to approach the validation? It would be nice to be able to balance the user validation feedback with what's being stored in the database (i.e. around storing `0's`), but I'm not sure what the best solution is.

acholyn:
I think if they expect a leading 0, we should allow this with a text field because if they find they're not able to input it we'll probably get a number of complaints that they can't input it in the same format as they've been given. Slightly unclear what the highest possible number is, is it still 999 since 3 digits and no leading 0 above 10?

as for the _interesting_ validation, the admin username field has multifaceted validation, so we could do something similar and flag the 
> numbers <10 need to start with a 0 but >=10 can't
?

PR #393, opened 2025-11-24
added UI/UX for picking project member roles

jhughes982:
## Works towards #10 

- This PR adds in the UI/UX for assigning additional project members and the environment specific roles,
- I have added some template roles for the DSH environment so I could start thinking how to make this flexible when we add future environments. Some of the DSH details might not be quite right, so let me know if this needs adjusting.
- There is a bit of repetition in the defining of the member roles between the frontend and api.web.yaml, but we may have to live with it as I don't think you can directly use the values of the enums from api.web.yaml in other files?
-  This will be the last frontend PR before I start incorporating the projects into the backend and storing the data in the DB. So perhaps we can collectively have a last look through all the implementations so far and let me know if there are any other details or edge cases that have been missed. 
  - For instance, have we captured all the necessary roles for TRE? 
  - Are there any other project attributes we need to consider at any level

EDIT:
- Based on internal discussions, I have temporarily removed the DSH roles until further decisions have been made.

Example of the UI for adding additional researchers and their roles
Screenshot 2025-11-24 at 17 00 14

### Checklist

### DIFF
--- a/web/src/components/projects/CreateProjectForm.tsx
+++ b/web/src/components/projects/CreateProjectForm.tsx
@@ -45,7 +45,7 @@ const TRE_ROLE_DESCRIPTIONS: Record<ProjectTreRoleName, string> = {
   egresser: "Can download data from the TRE environment after approval",
   egress_requester: "Can request data to be downloaded from the TRE",
   egress_checker: "Can review and approve egress requests from other users",
-  trusted_egresser: "Can download data from the TRE environment without requiring approval",
+  trusted_egresser: "Can download potentially sensitive data from the TRE environment to trusted locations",
 };
 
 const ROLE_LABELS: Record<AnyProjectRoleName, string> = {


web/src/components/projects/CreateProjectForm.tsx
```diff
@@ -436,81 +474,144 @@ export default function CreateProjectForm({ approvedStudies, handleProjectCreate
...
+                                },
+                              }}
+                              render={({ field: usernameField, fieldState }) => (
+                                <div className={styles["username-input-wrapper"]}>
+                                  <div>
+                                    <input
+                                      {...usernameField}
+                                      id={`researcher-${index}`}
+                                      type="text"
+                                      placeholder="Valid username"
```
t-young31: I don't think this is a useful placeholder. at UCL what my "username" is depends on the context. I vote 
```suggestion
                                      placeholder="ccaxyz"
```
but other suggestions welcome

web/src/components/projects/CreateProjectForm.tsx
```diff
@@ -436,81 +474,144 @@ export default function CreateProjectForm({ approvedStudies, handleProjectCreate
...
                     variant="secondary"
                     size="small"
-                    onClick={() => appendResearcher({ value: "" })}
+                    onClick={() => appendResearcher({ username: "", roles: [] })}
                   >
                     Add Researcher
                   </Button>
                 </fieldset>
-                <HelperText>Optionally add approved researchers to this project</HelperText>
+                <HelperText>Optionally add approved researchers with their roles to this project</HelperText>
```
t-young31: to be an uploader you don't need to be an approved researcher

t-young31 (commented):
like the UX

web/src/components/projects/CreateProjectForm.tsx
acholyn: is there any requirement to be staff for any of the roles?

jhughes982:
I think this is ready to be merged now. There are some ongoing discussions in this PR which I will come back to as we gain more knowledge and make more decisions about how this will work (e.g. regarding DSH).

web/src/components/projects/CreateProjectForm.tsx
```diff
@@ -3,28 +3,73 @@ import { Controller, useFieldArray, useForm } from "react-hook-form";
...
+  trusted_egresser: "Trusted Egresser",
+};
+
+const TRE_ROLE_DESCRIPTIONS: Record<ProjectTreRoleName, string> = {
+  desktop_user: "Can access the desktop environment and work with data within the TRE",
+  ingresser: "Can upload data into the TRE environment",
+  egresser: "Can download data from the TRE environment after approval",
+  egress_requester: "Can request data to be downloaded from the TRE",
+  egress_checker: "Can review and approve egress requests from other users",
+  trusted_egresser: "Can download potentially sensitive data from the TRE environment to trusted locations",
```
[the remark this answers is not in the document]
jhughes982: I don't feel strongly about this one. "download" is probably more universal, but it depends on the audience you are targeting I guess. Happy to implement it either way.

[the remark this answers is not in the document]
jhughes982: I'll merge as-is for now and we can revisit if anyone feels strongly about the wording 👍

PR #400, opened 2025-12-02
Projects backend

jhughes982:
## Works towards #10 

- This PR adds the backend handler and service to validate and store a TRE project along with its associated assets and project members,
- A bit of frontend abstraction was also added to handle the project name validation,
- There are still a few remaining questions around the implementation, and I have left a few todo comments in the code for further discussion. e.g.:
  - One requirement is that study admins should be able to make a project for a study they belong to, but I don't think the RBAC is set up for this yet? But maybe I missed something here and it's already set up?
  - I have temporarily removed the ability to save a project in draft mode as there are further conversations to be had around this and it can be added later.
  - The admin views around projects will be added in follow up PR's.
  - Once this has been signed off, I'll add some tests.

### Checklist

- [ ] Documentation has been updated

PR #407, opened 2025-12-05
inherit project owner role from study owner role

t-young31:
Works towards #10 - resolves my [comment](https://github.com/ucl-arc-tre/portal/pull/400#discussion_r2580700783) in the backend projects PR

### Checklist

### DIFF
--- a/internal/rbac/main.go
+++ b/internal/rbac/main.go
@@ -100,17 +100,17 @@ func StudyIDsWithRole(user types.User, studyRoleName StudyRoleName) ([]uuid.UUID
 // Add a project TRE owner role and add the role binding to
 // the parent study role so study owners inherit a project owner role
 func AddProjectTreOwnerRole(studyId uuid.UUID, projectId uuid.UUID) (bool, error) {
-	roleName := makeProjectOwnerRole(projectId).RoleName()
+	projectOwnerRole := makeProjectOwnerRole(projectId)
 	policy := Policy{
-		RoleName: roleName,
+		RoleName: projectOwnerRole.RoleName(),
 		Action:   "*",
 		Resource: fmt.Sprintf("/projects/tre/%v", projectId),
 	}
 	if _, err := addPolicy(enforcer, policy); err != nil {
 		return false, err
 	}
-	studyRole := StudyRole{StudyID: studyId, Name: StudyOwner}
-	return AddChildRole(studyRole.RoleName(), roleName)
+	studyOwnerRole := makeStudyOwnerRole(studyId)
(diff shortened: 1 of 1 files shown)


internal/rbac/main.go
```diff
@@ -143,8 +152,11 @@ func Roles(user types.User) ([]RoleName, error) {
 
 // Does the user have a role?
 func HasRole(user types.User, role RoleName) (bool, error) {
-	hasRole, err := enforcer.HasRoleForUser(user.ID.String(), string(role))
```
t-young31: `HasRoleForUser` doesn't include implicit ones, so switch to `slices.Contains` on the implicit ones (i.e. include inherited)

internal/service/projects/main.go
```diff
@@ -253,12 +253,13 @@ func (s *Service) CreateProjectTRE(ctx context.Context, creator types.User, stud
...
-		return types.NewErrFromGorm(err, "failed to commit create project transaction")
+	if _, err := rbac.AddProjectTreOwnerRole(studyUUID, project.ID); err != nil {
+		tx.Rollback()
+		return err
 	}
 
-	if _, err := rbac.AddProjectTreOwnerRole(creator, project.ID); err != nil {
-		return err
+	if err := tx.Commit().Error; err != nil {
+		return types.NewErrFromGorm(err, "failed to commit create project transaction")
```
t-young31: switched the order here to rollback the transaction if the role create fails because without it they'll be a project that no one can see. ok @jhughes982?

jhughes982:
> sorry for the overlap! should have asked what you were working on 😅

No worries!

internal/rbac/main.go
```diff
@@ -101,7 +109,8 @@ func AddProjectTreOwnerRole(user types.User, projectId uuid.UUID) (bool, error)
 	if _, err := addPolicy(enforcer, policy); err != nil {
 		return false, err
 	}
-	return AddRole(user, roleName)
+	studyRole := StudyRole{StudyID: studyId, Name: StudyOwner}
+	return AddChildRole(studyRole.RoleName(), roleName)
```
jhughes982: Is it possible to reuse `makeStudyOwnerRole` here? So something like:

```
projectOwnerRole := makeProjectOwnerRole(projectId).RoleName()
studyOwnerRole := makeStudyOwnerRole(studyId).RoleName()
...
return AddChildRole(studyOwnerRole, projectOwnerRole)
```

internal/service/projects/main.go
jhughes982: Looks good!

PR #426, opened 2025-12-12
moving methods into their own files

jhughes982:
- A quick housekeeping PR related to #427 
- This PR moves the asset related handler methods into `assets.go`
- Based on a [previous discussion](https://github.com/ucl-arc-tre/portal/pull/414#discussion_r2606890541)
- Opening as draft for now to get feedback. Does this feel tidier to everyone? If so, I could also move contract-related methods into its own `contracts.go` file to make `studies.go` a bit leaner.

If we feel this is not needed, happy to close.

### Checklist

### DIFF
--- a/internal/handler/web/assets.go
+++ b/internal/handler/web/assets.go
@@ -10,6 +10,8 @@ import (
 	"github.com/ucl-arc-tre/portal/internal/types"
 )
 
+// Helper functions
+
 func assetToOpenApiAsset(asset types.Asset) openapi.Asset {
 	return openapi.Asset{
 		Id:                   asset.ID.String(),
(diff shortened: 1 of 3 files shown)


jhughes982:
I decided to move the contracts-related methods so you can evaluate whether you like it this way or not. Can revert it if desired.

t-young31 (approved):
I'm weakly against this, because the namespacing is now no longer correct – `/studies/{studyId}/assets` is now `assets.go` but if you and @acholyn prefer then go for it

acholyn (approved):
like this, will be interesting to see (re. Tom's comment) if new devs find it confusing or not

PR #434, opened 2025-12-17
misc UI updates

### LINKED ISSUE #404: UI suggestions
## Assorted suggestions 

Based on end to end Profile -> Study - > Assets -> Contracts -> Mark ready for Review -> Study Review by IG Admin -> Request Project

---

- [ ] **Profile**: when you submit a valid training certificate and successfully complete your profile, you just get this component pop up. And `approved-staff-researcher` , `approved-researcher` etc. appear in the Roles section. But it would be nice to     have a          more explicit "Your Profile is Complete!" message somewhere?:
Image

- [ ] **Marking Study Ready for Review**: once my Study becomes ready for review (i.e. has signed study agreement plus at least 1 asset), while the `Mark Ready for Review` Button gets rendered, the Status remains `INCOMPLETE` and it's not hugely obvious that the Study is ready for review now. There is enough on the screen that I might not notice the Mark Ready for Review button has rendered. Could add a bounce effect on the button like we do with the green arrows?
Image

- [ ] **Study Review by Admin**: when reviewing a Study, this asset callout is still rendered - we could hide it for the admins since this is a callout for users?
Image

- [ ] **When IG Admin approves a Study**: when I hit approve, I get a "Feedback updated successfully" callout. Is that the wrong message?
Image

- [ ] **When the IAO looks at their newly approved Study**: the `Mark Ready for Review` button is still rendered:
Image

t-young31:
### Checklist


t-young31:
yeah – it's not. I couldn't come up with a nice way to do it, nor do I think it's completely necessary. thoughts welcome 
cc: @finleybacon

finleybacon:
> I couldn't come up with a nice way to do it, nor do I think it's completely necessary. thoughts welcome

it may not be an issue for others, we can always come back to it if we get feedback from users in testing that the next step is not clear. if others do think it's unclear, an option could be a tooltip text when you hover over the incomplete status saying (You need to mark this Study ready for review). Happy to leave it for now though

Issue #11, opened 2025-04-04
Project management


jackhindleyucl:
## Description

**As an** information asset owner or administrator
**I want to** manage my Project
**So that** we can continue to collaborate on our research in the TRE

## Acceptance criteria

- [ ] ~~View all projects from the top level Project page (searchable/filterable list e.g. [here](https://primereact.org/datatable/)), select specific object to go into manage project page -> to investigate (James, Finley)~~ Moved into #439 
- [x] Edit the project from the manage project page
- [x] Add Approved Researcher to the Project with certain permissions
- [x] Add certain permissions to an existing Project user
- [x] Remove a user from the Project
- [x] Revoke certain permissions from an existing Project user
- [x] "Delete" a Project - soft delete, no one can see it
- ~~[ ] "Delete" all Projects under a Study~~ Moved into #454 
- ~~[ ] "Archive" a Project - specific status, read-only access for IAO only, the data can't be accessed without a request from the IAO~~ Moved into #446 
- [x] Remove a user from all Projects I am owner/administrator of
- ~~[ ] (in case someone has left but not told everyone they should) Alert other Study owners/administrators that a user who has access to their assets has left~~ Moved into #456 
- ~~[ ] Users must be added in relation to one or more Information Assets~~
- ~~[ ] Any External user added must be added in relation to one or more Information Assets with a linked Contract, and one or more of those Contracts must be selected to add the user~~ Moved into #457 
- ~~[ ] Remind project owners to review access and record responses (ISMS compliance)~~ Moved into #456 
- [x] Asset addition + deletion from a project with a tier lower than the environment

## Dependencies

- #10 

Additional: 
- is there a distinction between "Archived" and "Deleted"
  - yes, checklist updated

jhughes982:
> Asset addition + deletion from a project with a tier lower than the environment

For this requirement, do we want to maintain a record of all the asset deletions and additions over multiple rows for a single asset, or is it ok to soft delete the asset from a project and then just set the `deleted_at` field back to `Null` if a user adds the asset back in at a later time?

t-young31:
I don't _think_ the history is important for project-asset linkage, however it may be for project role bindings. if the latter then might as well be consistent and allow multiple rows everywhere (inc. #410)

@gsvarovsky / @finleybacon what do you think?

jhughes982:
> tldr; do undelete and add history later if we need it'

Sounds good

jhughes982:
[...]

PR #452, opened 2026-01-15
allowing project delete

jhughes982:
## Works towards #11 

- This PR adds the delete functionality to projects,
- Projects will be soft deleted along with all associated project attributes across other tables,

Screenshot 2026-01-15 at 16 07 16
Screenshot 2026-01-15 at 16 07 47

### Checklist

### DIFF
--- a/web/src/components/projects/ProjectCardsList.module.css
+++ b/web/src/components/projects/ProjectCardsList.module.css
@@ -82,6 +82,36 @@
   align-items: center;
 }
 
+.delete-button {
+  background-color: #dc3545 !important;
+  border-color: #dc3545 !important;
+  color: white !important;
+}
+
+.delete-button:hover:not(:disabled) {
+  background-color: #c82333 !important;
+  border-color: #bd2130 !important;
+}
+
+.delete-button:disabled {
+  opacity: 0.6 !important;
+}
+
+.delete-button-confirm {
+  background-color: #dc3545 !important;
+  border-color: #dc3545 !important;
+  color: white !important;
+}
+
+.delete-button-confirm:hover:not(:disabled) {
+  background-color: #c82333 !important;
+  border-color: #bd2130 !important;
+}
+
+.delete-button-confirm:disabled {
+  opacity: 0.6 !important;
+}
+
 .delete-dialog {
   padding: 1rem 0.5rem;
   width: 100%;
--- a/web/src/components/projects/ProjectCardsList.tsx
+++ b/web/src/components/projects/ProjectCardsList.tsx
@@ -103,6 +103,7 @@ export default function ProjectCardsList(props: Props) {
                     size="small"
                     variant="secondary"
                     disabled={deletingProjectId === project.id}
+                    className={styles["delete-button"]}
                   >
                     {deletingProjectId === project.id ? "Deleting..." : "Delete"}
                   </Button>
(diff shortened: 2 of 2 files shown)


internal/service/projects/main.go
```diff
@@ -369,6 +369,53 @@ func (s *Service) UpdateProjectTRE(projectTRE *types.ProjectTRE, projectUpdateDa
...
+	// Retrieve the ProjectTRE
+	var projectTRE types.ProjectTRE
+	err := tx.Where("project_id = ?", projectId).First(&projectTRE).Error
+	if err != nil {
+		tx.Rollback()
+		return types.NewErrFromGorm(err, "failed to find project TRE")
+	}
+
+	// Soft delete all ProjectTRERoleBindings for this project
+	err = tx.Where("project_tre_id = ?", 123).Delete(&types.ProjectTRERoleBinding{}).Error
```
t-young31: `123`

web/src/components/projects/ProjectCardsList.tsx
```diff
@@ -49,10 +97,52 @@ export default function ProjectCardsList(props: Props) {
...
       </div>
+
+      {showDeleteConfirm && projectToDelete && (
+        <Dialog setDialogOpen={handleCancelDelete}>
+          <div className={styles["delete-dialog"]}>
+            <h2>Delete Project</h2>
+            <p>
+              Are you sure you want to delete project <strong>{projectToDelete.name}</strong>?
+            </p>
+            <p>This action will soft delete the project and all its associated data including:</p>
```
t-young31: I'm not sure if users will know what 'soft' deleting is

internal/service/projects/main.go
```diff
@@ -369,6 +369,53 @@ func (s *Service) UpdateProjectTRE(projectTRE *types.ProjectTRE, projectUpdateDa
 	return nil
 }
 
+func (s *Service) DeleteProjectTRE(projectId uuid.UUID) error {
+	tx := s.db.Begin()
+	defer graceful.RollbackTransactionOnPanic(tx)
+
+	// Retrieve the ProjectTRE
+	var projectTRE types.ProjectTRE
+	err := tx.Where("project_id = ?", projectId).First(&projectTRE).Error
```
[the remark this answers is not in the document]
t-young31: yeah definitely clear, just suggesting a (potentially premature!) optimisation. only the `projectTRE.ID` is needed, so could be just a select on the id, rather than the whole object. probably needless – feel free to ignore

web/src/components/projects/ProjectCardsList.tsx
```diff
@@ -49,10 +97,56 @@ export default function ProjectCardsList(props: Props) {
...
+            <p>This action will delete the project and remove the links to its associated data including:</p>
+            <ul>
+              <li>Project members and their roles</li>
+              <li>Linked assets</li>
+              <li>Project configuration</li>
+            </ul>
+            <p>
+              Note that the original data associated with this project will not be deleted, only the link between the
+              data and the project will be affected.
+            </p>
```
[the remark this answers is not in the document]
finleybacon: yep i think regardless of whether we actually wipe the project storage or just kill the desktops and keep the volumes for some extended window, we only want people to click delete if they really don't care about the data anymore. in some services there may be an archive function for people who want it.

so delete should really mean delete (in the mind of the user) i think

web/src/components/projects/ProjectCardsList.module.css
```diff
@@ -90,4 +138,16 @@
...
+
+  .delete-dialog {
+    max-width: 100%;
+  }
+
+  .delete-actions {
+    flex-direction: column-reverse;
+  }
+
+  .delete-actions button {
```
acholyn: could we make the delete button red in the dialog? That feels sufficiently common to me

PR #469, opened 2026-01-28
init metrics

### LINKED ISSUE #345: collection & display of metrics
OMG would like some metrics such as 
- percentage of users who are approved researchers
- percentage of those who have valid training
...

We could run these every x days/months, or we could do it as a live view available to admins, which could be checked at any time (and could give us an opportunity for fun graphs)

t-young31:
## Resolves #345

Uses [`react-google-charts`](https://www.react-google-charts.com/) to add a first pass at metrics that IG will be interested in. Much fun was had

Screenshot 2026-01-29 at 10 58 24

 
### Checklist

### DIFF
--- a/web/src/components/metrics/Metrics.tsx
+++ b/web/src/components/metrics/Metrics.tsx
@@ -44,7 +44,7 @@ export default function Metrics() {
         chartType="PieChart"
         data={data}
         options={{
-          title: "Number of users approved researchers with valid and invalid training",
+          title: "Number of approved researchers with valid and invalid training",
         }}
         width={"100%"}
         height={"400px"}


internal/service/users/training.go
```diff
@@ -137,6 +138,28 @@ func (s *Service) NHSDTrainingExpiresAt(user types.User) (*time.Time, error) {
...
+}
+
+func (s *Service) NumApprovedResearchersExpiredTraining() (int, error) {
+	var count int64
+	result := s.db.Model(&types.User{}).
+		Joins("INNER JOIN user_agreement_confirmations ON users.id = user_id").
+		Joins("INNER JOIN agreements ON user_agreement_confirmations.agreement_id = agreements.id").
+		Joins("INNER JOIN user_training_records ON users.id = user_training_records.user_id").
+		Where("agreements.type = ? AND user_training_records.kind = ? AND user_training_records.completed_at < now() - interval '1 year'", agreements.ApprovedResearcherType, types.TrainingKindNHSD).
+		Count(&count)
```
t-young31: I will undo this copy paste

acholyn: is it better to do these calcs on the backend rather than the frontend? @jhughes982 any thoughts on this?

web/src/components/people/Metrics.tsx
```diff
@@ -0,0 +1,45 @@
...
+    ["Approved valid", num_valid_training],
+    ["Approved expired", num_expired_training],
+    ["Other", num_other],
+  ];
+
+  if (isLoading) return null;
+
+  return (
+    <>
+      <Chart chartType="PieChart" data={data} options={{}} width={"100%"} height={"400px"} />
```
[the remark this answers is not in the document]
finleybacon: ok retracted my earlier vote, discussed with Tom yesterday, and we're working with the status quo which says that expired training does **not** revoke your Approved Researcher role, instead you become something like an "expired Approved Researcher". 

So the graph in the PR description is what we need, but we need one of them per environment, so they can inform claims like "95% of DSH users are approved researchers with valid training" or "100% of ARC TRE users are approved researchers with valid training"

Issue #476, opened 2026-02-05
rework study page


acholyn:
Since moving contracts from an Asset attachment to a Study attachment, along with the inclusion of Projects, we should have a nicer interface for the Study page such that a user can see:

- [x] summary of study (StudyDetails)
  - [ ] + summary of number of projects, assets and contracts [& number of users?]
- [ ] the Projects (as tab)
- [x] the Contracts (as tab)
- [x] the Assets (as tab)
- [ ] the users? [IAO/IAA only] (as tab) - nice to have, more important at Project level

cc @jhughes982 @finleybacon thoughts?

acholyn:
having some more consistency between the assets and projects components + their lists would probably help; projects currently doesn't accept a study arg to only get projects for one study, see #495

jhughes982:
Regarding the `Projects` navbar item on the left. Do we want this to disappear in favour of moving all the projects functionality under the projects tab item (e.g. seeing list of projects, creating projects etc)?

acholyn:
> Regarding the `Projects` navbar item on the left. Do we want this to disappear in favour of moving all the projects functionality under the projects tab item (e.g. seeing list of projects, creating projects etc)?

nope, we keep the projects page in the navbar still; that's for _all_ projects for a user
the tab on the study page is for projects within that specific study

I think creating a new project should be on the project page and on the study page is kind of view-only, but if users find it annoying to go to the project page to create a new project within a study then we can always add a "new project" button on the project tab on the study page

jhughes982:
Ok, I anticipate the multiple project areas might be a source of confusion for users. But perhaps we can mitigate this with clear text and labelling.

Edit:
I personally lean towards mimicking the current behaviour of assets and contracts for projects. This way users always have one flow for doing everything, i.e. they access their projects tab and create a project from there just like they do for other study entities.

Unless users also need to see all projects in one big list for some reason (or maybe I'm missing some other context)? If not, then I think users accessing projects via the tab should be fine right?

jhughes982:
@t-young31 do you have any thoughts on the above? Happy to stick to the current plan of having projects in 2 different areas if I'm in the minority.

acholyn:
> Ok, I anticipate the multiple project areas might be a source of confusion for users. But perhaps we can mitigate this with clear text and labelling.
> 
[...]

Issue #435, opened 2025-12-18
Link Assets to Contracts


finleybacon:
## Description

**As a** IAO/IAA
**I want to** know which contract each of my assets is governed by (if they are governed by a contract)
**So that** when one of my contracts expires, I know exactly which asset(s) are impacted (and which Project(s) they are being used in). 

## Acceptance criteria

- [ ] Criteria 1
- [ ] ...

## Dependencies

- #ISSUE_NUMBER

see also: #378

PR #480, opened 2026-02-06
UI for contract flow + asset linkage

### LINKED ISSUE #436: Change Asset Contracts -> Study Contracts
i.e. change Contract struct to have Study ID instead of AssetID

This means contracts will belong to Studies, not Assets. Introducing Asset-Contract links is a possible future feature, per the user Story in issue #435 

### Steps

- [ ] Change the struct/db schema
- [ ] Change the Study flow to make asset creation and study creation distinct
- [ ] Update docs (esp ERD in HLD)

acholyn:
## Closes #436 
also related to #435 and work towards #476

### Checklist

- simplified study page for displaying details, assets and contracts, with the expectation of this to be expanded for projects (related to #476 ), including new StudyOverview component
- added contracts to study page
image
image

- added assets field to contract form
image

- updated tests that were failing due to changes in how assets and contracts and linked
- added note in docs for failing cypress tests due to auth timeout

### DIFF
--- a/web/src/components/studies/StudyDetails.tsx
+++ b/web/src/components/studies/StudyDetails.tsx
@@ -254,7 +254,10 @@ export default function StudyDetails(props: StudyDetailsProps) {
       {(studyStepsCompleted || isIGOpsStaff) && (
         <>
           <div className={`${styles["tab-content"]} ${tab === "assets" ? styles.active : ""}`}>
-            <Assets studyId={study.id} canModify={isStudyOwnerOrAdmin} setNumAssets={setNumAssets} />
+            {tab === "assets" && (
+              // we want to check the assets have required contracts more regularly
+              <Assets studyId={study.id} canModify={isStudyOwnerOrAdmin} setNumAssets={setNumAssets} />
+            )}
           </div>
 
           <div className={`${styles["tab-content"]} ${tab === "contracts" ? styles.active : ""}`}>


web/src/components/contracts/ContractUploadForm.tsx
```diff
@@ -10,22 +10,24 @@ import {
...
+import { HelperText, Label } from "../shared/exports";
 
 type ContractFormData = {
   organisationSignatory: string;
   thirdPartyName: string;
   status: "proposed" | "active" | "expired";
   startDate: string;
   expiryDate: string;
-  assetIds: string[];
+  assetIds: { value: string }[];
```
t-young31: `assetIds` should be a list of strings, imo. (`assets` would be a list of objects)

web/src/components/contracts/ContractUploadForm.tsx
```diff
@@ -44,6 +46,11 @@ export default function ContractUploadModal({
   const [uploading, setUploading] = useState(false);
   const [error, setError] = useState<string | null>(null);
   const [uploadSuccess, setUploadSuccess] = useState(false);
+  const [assetIds, setAssetIds] = useState<string[]>(throughAsset ? [throughAsset.id] : []);
```
t-young31: I'm confused by the split of state between the forms `assetIds` and the react state `assetIds`

web/src/components/contracts/ContractUploadForm.tsx
t-young31: oki. call it `assets` then

web/cypress/e2e/pages/contract-management.cy.ts
```diff
@@ -1,348 +1,435 @@
...
+    // Submit the form
+    cy.contains("Update Contract").click();
+
+    cy.wait("@editContract");
+    cy.wait("@getStudyContractsWithSample");
+
+    // Verify the mock contract is displayed
+    cy.contains("sample-contract.pdf").should("be.visible");
+    cy.contains("Sample Organization").should("be.visible");
+  });
```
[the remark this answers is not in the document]
acholyn: note: will probably be superseded by #498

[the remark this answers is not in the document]
t-young31: could rm and add to https://github.com/ucl-arc-tre/portal/blob/2f7a5880d4e8851ab206270a10e7b3a32e32e87a/web/cypress/e2e/integration/studies/creation.cy.ts#L19 ?

web/src/components/contracts/ContractManagement.tsx
jhughes982: I linked a contract to an asset, but I can't see anything in the card UI's for either the study, asset, or contract that visually indicates a link (unless I start editing the contract again to see it). Is this intentional? 

I expected this information to be important to users, and the glossary diagram indicates the relationships thoroughly. If so, should we add more information to the cards to indicate the linkage?

jhughes982 (commented):
This is where I got to for today. I focussed mostly on the UI/UX so far. Tomorrow morning I will have a few hours to look through the code more thoroughlly.

web/cypress/fixtures/assets-with-sample-no-contracts.json
jhughes982: Are we not trying to [move away](https://github.com/ucl-arc-tre/portal/issues/498) from the use of fixtures?

web/src/styles/globals.css
```diff
@@ -90,6 +90,39 @@ dialog {
   margin-bottom: 16px;
 }
 
+.tab-collection {
+  width: 100%;
+  display: grid;
+  grid-template-columns: repeat(auto-fit, minmax(200px, 1fr));
+  position: relative;
+  margin-block: 1.8rem;
```
jhughes982: If `globals.css` starts to get too long you can split this out into tabs related it's own css file perhaps?

web/src/components/studies/StudyDetails.tsx
```diff
@@ -158,200 +171,102 @@ export default function StudyDetails(props: StudyDetailsProps) {
...
-            ))}
-        </div>
+      {(isIGOpsStaff || studyStepsCompleted) && (
+        <>
+          <div className={"tab-collection"}>
+            <Button
+              onClick={() => setTab("overview")}
+              variant="secondary"
+              className={`tab ${tab === "overview" ? "active" : ""}`}
+            >
```
jhughes982: To make it super clear, I would be tempted to call this tab groupd `Study Overview` or even just `Study`. I don't feel too strongly about it though.

web/src/components/contracts/ContractManagement.tsx
```diff
@@ -129,7 +149,7 @@ export default function ContractManagement({ study, asset, canModify }: Contract
       {canModify && (
         <ContractUploadForm
           study={study}
-          assetIds={asset ? [asset.id] : []} //TODO: Add support for multiple assets
+          throughAsset={asset}
           isOpen={showUploadModal}
           onClose={() => {
             setShowUploadModal(false);
```
jhughes982: There are a few issues with this style of implementing a conditional component.
1. The component may be unnecessarily rendered even when the user has not clicked to open the contract form.
2. It's a bit misleading in that on first glance you might not realise that this is meant to be a conditionally rendered component, you might guess it from the props but it's indirect which means you have to open the component to really see whats going on.

Is there a way to make it so the conditional logic to render the component is inside `ContractManagement.tsx`?

web/cypress/e2e/pages/contract-management.cy.ts
```diff
@@ -1,348 +1,190 @@
...
+
+    //study should be valid from initial setup
+    cy.wait(["@getStudyById"]);
+
+    cy.wait("@getStudyAgreementText");
+    cy.wait("@getStudyAgreementsConfirmed");
+    cy.wait("@getAssetsWithSampleNoContracts");
+
+    // Should show empty contract state
+    cy.wait(1000);
```
jhughes982: What's the use of `cy.wait(1000)` here? It's not usually a recommended way of circumventing test failures (e.g. due to a lack of data). You would usually want to use something like `cy.wait("@getAssetsWithSampleNoContracts");`

web/src/components/contracts/ContractUploadForm.tsx
jhughes982: I also find the use of `const [assetIds, setAssetIds] = useState<string[]>(throughAsset ? [throughAsset.id] : []);` confusing. 

What is the context around the use of `throughAsset`? Do you have a test that simulates what the expected behaviour is here? Since I think the component logic never actually passes in an `asset` via props (unless I'm mistaken?) I can't test this myself to see what should/does happen.

web/cypress/support/commands.ts
```diff
@@ -640,6 +664,45 @@ Cypress.Commands.add("mockAssetAccess", () => {
...
+        id: "contract-2",
+        organisation_signatory: "Sample Organisation 2",
+        start_date: "2026-02-22",
+        status: "active",
+        study_id: "123456789",
+        third_party_name: "Sample Third Party",
+        updated_at: "2026-02-13T15:59:21Z",
+      },
+    ],
+  }).as("getContractsForAsset");
```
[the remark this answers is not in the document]
acholyn: just followed the pattern for the other access mock 🙈

web/cypress/e2e/pages/contract-management.cy.ts
acholyn: I've noticed that when there's a number of mocks to do, at least locally, this helps with the timing

web/src/components/contracts/ContractManagement.tsx
```diff
@@ -129,7 +149,7 @@ export default function ContractManagement({ study, asset, canModify }: Contract
       {canModify && (
         <ContractUploadForm
           study={study}
-          assetIds={asset ? [asset.id] : []} //TODO: Add support for multiple assets
+          throughAsset={asset}
```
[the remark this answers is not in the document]
acholyn: the asset through which the contract form has been triggered. may be more worth me stripping this out in this pr and just building in the linkage form in a separate pr

web/src/components/studies/StudyDetails.tsx
```diff
@@ -89,14 +88,28 @@ const calculateRiskScore = async (study: Study) => {
...
+  } = props;
   const [riskScore, setRiskScore] = useState(0);
   const [riskScoreLoading, setRiskScoreLoading] = useState(false);
   const [feedback, setFeedback] = useState("");
   const [error, setError] = useState<string | null>(null);
   const [approvalStatus, setApprovalStatus] = useState<ApprovalStatus | undefined>(undefined);
 
+  const [tab, setTab] = useState("overview");
+  const [numAssets, setNumAssets] = useState(0);
+  const [numContracts, setNumContracts] = useState(0);
```
t-young31: think this should default to `undefined` and only show if set to a number (showing `0` before loaded is kinda misleading)

PR #485, opened 2026-02-10
change name regex to enforce a space

### LINKED ISSUE #484: Update chosen name field validation to enforce >1 word
## Description

**As a** user
**I want to** not accidentially set my name to just my first name
**So that** it's more likely to match my training certificate

https://github.com/ucl-arc-tre/portal/blob/6cc07a16614bef8374086dadad520c11329f75bd/web/src/components/profile/approved-researcher-components/ChosenName.tsx#L78

- [ ] update form to say "full name (first name + surname)"
- [ ] stretch: auto populate from entra, but make it editable

t-young31:
### Checklist

### DIFF
--- a/web/src/components/profile/approved-researcher-components/ChosenName.tsx
+++ b/web/src/components/profile/approved-researcher-components/ChosenName.tsx
@@ -8,11 +8,11 @@ import { Alert, AlertMessage, Input } from "@/components/shared/exports";
 
 type ProfileChosenNameProps = {
   currentName?: string;
-  setName: (name: string) => void;
+  setChosenName: (name: string) => void;
 };
 
 export default function ProfileChosenName(props: ProfileChosenNameProps) {
-  const { currentName, setName: setName } = props;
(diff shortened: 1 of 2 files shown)


web/src/components/profile/approved-researcher-components/ChosenName.tsx
```diff
@@ -1,48 +1,38 @@
...
-  ssr: false,
-});
-const Input = dynamic(() => import("uikit-react-public").then((mod) => mod.Input), {
-  ssr: false,
-});
+import { Alert, AlertMessage, Input } from "@/components/shared/exports";
 
 type ProfileChosenNameProps = {
   currentName?: string;
-  setChosenName: (name: string) => void;
```
acholyn: nit: feel like this could've been left as setChosenName since we still use ChosenName in other places 🤷‍♀️

t-young31: fair 66c0c3b4

PR #503, opened 2026-02-18
Delete relationships png

### LINKED ISSUE #502: Source of truth for app entity relationships
- We currently have two diagrams in the repo that explain the relationship between studies, assets, and contracts:
  - https://github.com/ucl-arc-tre/portal/blob/main/docs/media/entity%20diagram.drawio.png
  - https://github.com/ucl-arc-tre/portal/blob/main/web/public/entity_diagram.drawio.svg

- Which one is the correct one? Or are both needed?

## Requirements
- [ ] Remove any diagrams that are outdated

@acholyn @t-young31 @finleybacon

acholyn:
resolves #502

### DIFF
--- a/docs/media/entity_diagram.drawio.svg
+++ b/docs/media/entity_diagram.drawio.svg
@@ -1,1 +1,1 @@
-./web/public/entity_diagram.drawio.svg
+../../web/public/entity_diagram.drawio.svg


jhughes982:
> @finleybacon did you update the drawio file as well or just output the image?

It looks the same as the `.png` version so should probably be deleted.

t-young31 (approved):

Issue #526, opened 2026-02-27
Add `last_signoff` study attribute


jhughes982:
## Requirements
- [x] Newly created studies should have this column added
- [x] On study creation `last_signoff` is set to null
- [x] `last_signoff` gets set as the current datetime as soon as the study has been approved for the first time
- [ ] A migration needs to run to create the `last_signoff` column to all existing studies in staging and prod
- [ ] The migration script is removed once the column has been added to all existing studies on staging and prod

PR #527, opened 2026-02-27
added last_signoff migration

jhughes982:
## Works towards #526 

- This PR adds migration code to add the `last_signoff` study attribute to existing studies in staging and prod

### Checklist

### DIFF
--- a/internal/graceful/db.go
+++ b/internal/graceful/db.go
@@ -43,6 +43,7 @@ func InitDB() {
 	db.Exec(`CREATE EXTENSION IF NOT EXISTS "uuid-ossp";`)
 
 	migrateContractStudyIds(db)
+	migrateStudyLastSignoff(db)
 
 	if err := db.AutoMigrate(models...); err != nil {
 		panic(err)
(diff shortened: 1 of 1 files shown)


internal/graceful/db.go
```diff
@@ -110,6 +111,32 @@ func migrateContractStudyIds(db *gorm.DB) {
...
+
+	if err := migrator.AddColumn(&study, "last_signoff"); err != nil {
+		panic(err)
+	}
+
+	if err := db.Model(&study).
+		Where("approval_status = ?", "Approved").
+		Update("last_signoff", gorm.Expr("updated_at")).Error; err != nil {
+		panic(err)
+	}
```
jhughes982: This was just a thought for adding a `last_signoff` date to already existing approved studies. So if a study is already approved the `last_signoff` attribute gets set as the same date as the `updated_at` column. I'm not sure if this is acceptable or not though, so thoughts welcome. We could also just set them all to `null` until the later logic for this attribute is added in a future PR.

PR #529, opened 2026-02-28
stub gocron background tasks

t-young31:
Prerequisite for sending reminders etc. in #284

### Checklist

### DIFF
--- a/deploy/dev/config.sample.yaml
+++ b/deploy/dev/config.sample.yaml
@@ -1,7 +1,8 @@
 debug: true
 
 # Deploy-time configurable list of role bindings for identity
-# provided usernames
+# These allow you to log into the portal as different users with different permissions
+# Ask a senior member of the team to generate these for you
 admin_usernames: # Global portal admins
   - bob@example.com
 tre_ops_staff_usernames: # Trusted research environment staff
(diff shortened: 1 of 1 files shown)


internal/tasks/main_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+	calledFuncInManager := false
+	setCalled := func() error {
+		calledFuncInManager = true
+		return nil
+	}
+
+	manager := Manager{scheduler: scheduler}
+	manager.mustEvery(50*time.Millisecond, setCalled, "test")
+	manager.scheduler.Start()
```
acholyn: is this how I'd call the new jobs? eg. checkContractsExpiry.mustEvery(...)

internal/tasks/main.go
```diff
@@ -0,0 +1,82 @@
...
+// Shutdown the task manager. Errors are logged
+func (m *Manager) Shutdown() {
+	err := m.scheduler.Shutdown()
+	if err != nil {
+		log.Err(err).Msg("Failed to shutdown gocron scheduler")
+	}
+}
+
+func (m *Manager) mustEvery(delay time.Duration, f func() error, name string) {
+	job, err := m.scheduler.NewJob(
```
jhughes982: Personal preference would be to use a descriptive name rather than `f`. In this case it is clear whats going on, so I'll leave it to you whether you want to change this or not.

t-young31: [a3c48fb](https://github.com/ucl-arc-tre/portal/pull/529/commits/a3c48fb8a92a3c9a4bfc191586f3e987b93ab61d)

PR #554, opened 2026-03-08
disable studies in prod

### LINKED ISSUE #541: Strategy for releasing features
- On the studies page (and maybe other pages) there is a warning saying that the feature is still under development and any data might be lost.
- As a user, I probably would avoid using the service if that was the case (although I think some people are actually using the portal already).
- It might be worth discussing the strategy of releasing features.
- From my own perspective, I would prefer to be more cautious around feature releases until we are more confident with how they work.

Image

## Requirements
- [ ] Discuss with the team if we need to consider a strategy for feature releases.
- [ ] If the warning banner is no longer relevant then remove it (or at least the bit around data loss).

t-young31:
### Checklist


acholyn (approved):
🔒

PR #584, opened 2026-03-27
pagination

### LINKED ISSUE #580: study search results pagination
## Description

**As an** IG ops staff
**I want to** search studies and not overload the portal db
**So that** I can easily find the study I'm looking for without wasting resources

## Acceptance criteria

- [ ] paginate backend (quantity tbc)
- [ ] paginate frontend (quantity tbc)

acholyn:
### Checklist

- backend pagination, if no limit or offset given from frontend, they are set to 12 and 0 respectively. Each page has 12 results max
- ordered by `last_signoff`
image


web/src/components/studies/StudyCardsList.tsx
```diff
@@ -49,6 +66,20 @@ export default function StudyCardsList(props: Props) {
...
+          <Pagination
+            total={studies.length}
+            limit={limit}
+            offset={offset}
+            onPageChange={(newOffset) => setOffset(newOffset)}
+          >
+            <PaginationControls />
+            <PaginationInfo />
+          </Pagination>
+          {studies.length == 50 && <small>Please note these results have been limited to 50 items</small>}
```
t-young31: this needs to now use the value defined in the request. (basically: `50` should appear once in the backend and once in the frontend and nothing bad should happen if the frontend requests != 50)

t-young31 (commented):
backend LGTM. will leave the rest for @jhughes982

acholyn:
> A few observations:
> 
> 1. might be nice to have a `Clear search` to reset the state to all studies.
> 2. No matter how many studies are in the list, it looks like I can always click the `Next Page` button. I'd hide or disable the button if the number of studies hasn't reached the limit.
> 
> Screenshot 2026-04-01 at 15 44 26

the X should function as a clear

I'm open to suggestions for the second, as presently it will try to get the next page and if the next page is empty it then gets rid of the button and says you've reached the end of results, but you don't know that until you've tried getting the next page

unless you just meant like `if studies.length < studiesPerPage` hide next button

jhughes982:
> the X should function as a clear

Currently the 'x' clears the search term but it does not reset the study cards. There should be a way for the user to reset the state of the study cards without having to perform a search with a blank search term.

> I'm open to suggestions for the second, as presently it will try to get the next page and if the next page is empty it then gets rid of the button and says you've reached the end of results, but you don't know that until you've tried getting the next page
> 
> unless you just meant like `if studies.length < studiesPerPage` hide next button

Yes I do mean hide the next button if there are not enough studies.

jhughes982:
There is an edge case here in the scenario of having exactly the number of studies to fill the current pagination (e.g. 24, 36 etc.). In these cases when you reach the end of the studies list you can still see and click 'Next page'. This presents a confusing scenario to the user since they can click 'Next page' even though there are no more studies to show.

From my perspective, a better UX would be to also show how many studies there are in total (e.g. `Results: 25 - 36 of 100`) and when you reach the end of the list you just hide the 'Next page button' (or grey it out).

Here's some ideas. If that's extra work then I think at least showing some sort of total and preventing clicks on the 'Next page' button are worth implementing at this stage.

UX idea 1:
Screenshot 2026-04-09 at 12 56 37

UX idea 2:
Screenshot 2026-04-09 at 12 57 26

acholyn:
> There is an edge case here in the scenario of having exactly the number of studies to fill the current pagination (e.g. 24, 36 etc.). In these cases when you reach the end of the studies list you can still see and click 'Next page'. This presents a confusing scenario to the user since they can click 'Next page' even though there are no more studies to show.
> 
> From my perspective, a better UX would be to also show how many studies there are in total (e.g. `Results: 25 - 36 of 100`) and when you reach the end of the list you just hide the 'Next page button' (or grey it out).
> 
> Here's some ideas. If that's extra work then I think at least showing some sort of total and preventing clicks on the 'Next page' button are worth implementing at this stage.
> 
> UX idea 1: Screenshot 2026-04-09 at 12 56 37
> 
> UX idea 2: Screenshot 2026-04-09 at 12 57 26

due to how we get the results on the backend and limit them, we don't know what the total is

jhughes982:
Another edge case:

If I have already clicked through multiple pages and then make a search, it still maintains the old 'Showing results...' message and I can also click the 'Previous page' button which clears my search

Screenshot 2026-04-09 at 13 02 18

Edit:
Actually, I'm now not quite sure how the pagination is working. If I make a search and get back 12 results, I can still click the 'Next page' button and see results that were not returned in my original search.

Edit 2:
I have now observed that the same study is shown across multiple pages after clicking 'Next Page'. I think the pagination implementation might have a bug somewhere either on the frontend or backend.

Edit 3:
I also ran into a scenario where a study was not shown on any of the pages when clicking 'Next page' but it was returned when searching directly.

jhughes982:
> due to how we get the results on the backend and limit them, we don't know what the total is

I'm sure a simple query could return the total. Up to you if you want to explore it.

jhughes982:
This seems to be working better now, the bugs I spotted above look to be fixed.

However, there is still an edge case when there are exactly 12 studies. In this case the 'Next page' button still appears and I can click on it but nothing happens. Up to you if you want to dig deeper into that.

I'll have a look at the code now.

web/src/components/ui/Search.tsx
```diff
@@ -9,20 +10,39 @@ type Props = React.ComponentProps<typeof UCLSearch> & {
...
+  useEffect(() => {
+    if (onClear) {
+      const handleClear = () => {
+        onClear();
+      };
+      const clearButton = document.querySelector(
+        '[data-testid="ucl-uikit-search-clear-search-btn"]'
+      ) as HTMLButtonElement | null;
+
+      clearButton?.addEventListener("click", handleClear);
```
jhughes982, line 33: This isn't a typical React approach (i.e. using `querySelector` and `addEventListner`). However, it looks like this might be because you are implementing a workaround to interact with uikit.

My suggestion would be that if uikit isn't exposing the correct methods to fit our purpose then I would postpone using the component until it is ready. But if this is the most pragmatic approach and you have opened a uikit issue then this might just be the quickest way forward. I'll leave the final decision to you.

jhughes982:
There are two separate sections that mention the ordering/organisation. One of them is at the top of the list of studies and the other at the bottom, they also seem to contradict each other without further context.

Screenshot 2026-04-13 at 11 53 07

Screenshot 2026-04-13 at 11 53 19

web/src/components/studies/IGOpsStudies.tsx
```diff
@@ -124,7 +188,44 @@ export default function IGOpsStudies() {
...
+                  className={styles["prev-button"]}
+                  onClick={handleFetchPreviousPage}
+                >
+                  Previous Page
+                </Button>
+              )}
+              <small>
+                Showing studies {offset + 1} - {offset + studies.length}
+              </small>
+              {!(studies.length < studiesPerPage) && (
```
jhughes982:
```suggestion
              {studies.length >= studiesPerPage && (
```

I think that still follows what you're trying to do here, and in my opinion is easier to read. However, I'd still overall favour showing the total number of studies so the user intuitively knows where they are instead of having to click the 'next page' button again to return an empty list of studies. Up to you though.

web/src/components/studies/IGOpsStudies.tsx
jhughes982, line 1: There's three separate methods in this file that each call `getStudies` along with duplicated error handling logic. In addition the handling of the `offset` and query building logic feels more convoluted than it needs to be which makes the code harder to read. For instance, sometimes a `query` is passed to `getStudies` and sometimes it's not, but there's no clear indication as to why this difference exists.

Functionally the logic works, but it feels like it could be clearer with some refactoring. I'll leave it to you though if you want to spend extra time polishing this.

jhughes982 (approved):
The functionality seems to all work ok. It feels like the logic could be more clear and I'd be tempted to tweak some of the implementations (see my earlier comments) but since it all works I'll leave it to you whether you want to spend more time on this.

acholyn:
> There are two separate sections that mention the ordering/organisation. One of them is at the top of the list of studies and the other at the bottom, they also seem to contradict each other without further context.
> 
> Screenshot 2026-04-13 at 11 53 07 Screenshot 2026-04-13 at 11 53 19

Potentially for the top one I should say they're grouped, not organised, by status. Can also rm it or move it to the bottom. Was trying to be consistent with the use of text above since there's also some text for the pending tab

web/src/components/studies/IGOpsStudies.tsx
acholyn: number of returned studies shouldn't be more than the number per page but I guess that doesn't actually matter

web/src/components/studies/IGOpsStudies.tsx
acholyn: will leave for another PR, open to your suggestions or if you want to tackle it

PR #588, opened 2026-03-31
Study form components

jhughes982:
- Another round of refactoring.
- This PR separates the study form file (which had become very large) into smaller components that are easier to digest.
- I also did some more cleanup of redundant code, reducing the number of props being passed, and simplifying implementations.
- I also cleaned up the study search test which contained some invalid `it` nesting.

### Checklist


t-young31 (approved):
LGTM. one codex comment 

> Keep earlier-step fields mounted so submit still validates them — web/
    src/components/studies/study-form/StudyForm.tsx:121-133
    If a user enters an invalid optional value on step 2 (for example a malformed CAG, IRAS, or NHSE reference), then
    clicks through to step 3 and submits, react-hook-form will skip those validations because step 2 has been
    unmounted here. In the previous implementation the earlier fieldsets stayed mounted, so final submit still caught
    these errors. As written, invalid metadata can now be sent to the API whenever it was entered on a prior step.

which may, or may not, be correct. don't have the time to dig in right now

acholyn (approved):

PR #595, opened 2026-04-08
Package updates

jhughes982:
- This PR is an alternative to #590 
- All packages have been upgraded the same as #590 apart from typescript to prevent the build error observed in that PR. This is due to Typescript and Cypress being out of sync. Therefore we should perform the Typescript upgrade when it is back in sync with Cypress.
- This PR also makes the relevant Cypress code changes to comply with their recommended approach of handling env variables.

### Checklist


t-young31 (approved):
thanks @jhughes982 🙏🏼 🚀

PR #602, opened 2026-04-12
fix external users search

### LINKED ISSUE #599: User search not retrieving external users
When searching external users by either email or username (I think username = the user's home tenant UPN) on the /people page, no matches are found.

Maybe this is because the `FindUsernames` function in the entra controller is asking the UCL entra tenant to `Select: []string{"userPrincipalName"}`: and for external users, this is only returning the `john.doe_contoso.com#EXT#@yourtenant.onmicrosoft.com` UCL-tenant UPN, which doesn't give any matches for usernames in the portal database.

instead we probably want the user's home tenant UPN (this seems to be what the portal records as the "username" when they login in for first time), but not clear how we can get that from speaking to the UCL tenant.

t-young31:
## Resolves #599

- Queries the db before falling back to entra to find users with a case insensitive match
- _should_ fix the lookup of external users in prod
- Drive by: fixes some RBAC for users search

### Checklist


web/src/pages/people/index.tsx
```diff
@@ -77,7 +78,7 @@ export default function PeoplePage() {
     }
   };
 
-  if (!isAdmin && !isTreOpsStaff && !isIAO) {
+  if (!isAdmin && !isTreOpsStaff && !isIAO && !canSearch) {
```
acholyn:
```suggestion
  if (!canSearch) {
```

web/src/pages/people/index.tsx
```diff
@@ -130,7 +133,7 @@ export default function PeoplePage() {
         </Alert>
       )}
 
-      {!isAdmin && <Callout construction />}
+      {!canSearch && <Callout construction />}
```
acholyn: if they're not able to search, they can't view the page so don't think this is necessary. I think those who CAN search should still see the construction message until we've tested functionality
```suggestion
      {canSearch && <Callout construction />}
```

t-young31: > if they're not able to search, they can't view the page

erm, nope

> those who CAN search should still see the construction message

this is exactly the opposite of what it was

web/src/pages/people/index.tsx
t-young31: IAO's can't search but can invite externals, so would receive an incorrect 'you can't view this page' error

acholyn:
```suggestion
  if ( !isIAO && !canSearch) {
```

then? The repetition of admin and TREops is unnecessary at least

web/src/pages/people/index.tsx
acholyn: > > if they're not able to search, they can't view the page
> 
> erm, nope
> 
yeah, misread for the IAO; thought they could search as well

> > those who CAN search should still see the construction message
> 
> this is exactly the opposite of what it was

have revised my stance, think people who can search should know it's not flawless (yet)

web/src/pages/people/index.tsx
t-young31:
```suggestion
  if (!isIAO && !canSearch) {
```

web/src/pages/people/index.tsx
t-young31: none of the other functionality is flawless, so
```suggestion
      <Callout construction />
```
?

acholyn: touché
honestly maybe we should just have a persistent banner across the whole site 😆

acholyn (approved):

t-young31:
did you want to/have time to look at this before I merge @jhughes982? it would be good to ship this into prod on Wednesday

jhughes982:
Don't let me hold this up. I can look at the code changes after it's merged as tomorrow will be busy due to TI planning. Happy for you to merge this in 👍

PR #609, opened 2026-04-15
email notifs for training certificates

acholyn:
### Checklist

Training certs are checked every day, if they're due to expire in less than 30 days or have expired, a notif is sent if:
- already expired (<0)
- expires tomorrow (1)
- expires in 7 days
- expires in 14 days
- expires in 21 days
if not one of these, it shouldn't send an email [open to other ways of doing this]


internal/controller/entra/custom_mail.go
```diff
@@ -147,3 +147,23 @@ func (c *Controller) SendExpiryNotification(ctx context.Context, emails []string
 	notificationMsg := "Notification: Your contract in" + study.Title + " is due to expire soon"
 	return c.createCustomEmail(ctx, notificationMsg, emails, content)
 }
+
+func (c *Controller) SendTrainingExpiryNotification(ctx context.Context, emails []string, training types.UserTrainingRecord) error {
+	days := training.DaysUntilExpiry()
+	content := "You have a training certificate that is due to expire within"
+	if days != 1 && days > 0 {
+		if days == 21 || days == 14 || days == 7 {
```
t-young31: I don't think a malformated email should be sent if this is condition is not met

internal/tasks/expiry_checks.go
```diff
@@ -32,7 +32,38 @@ func (m *Manager) checkContractsExpiry() error {
...
+		return nil
+	}
+
+	ctx := context.Background()
+
+	trainingRecords := []types.UserTrainingRecord{}
+	result := m.db.Model(&types.UserTrainingRecord{}).Preload("User").Find(&trainingRecords)
+	if result.Error != nil {
+		return types.NewErrFromGorm(result.Error, "failed to get training records")
+	}
```
t-young31, line 60: I'd prefer if the manager had studies and user services to preserve the encapsulation (appreciate I should have said this in the previous notif PR)

internal/controller/entra/custom_mail.go
```diff
@@ -147,3 +147,23 @@ func (c *Controller) SendExpiryNotification(ctx context.Context, emails []string
 	notificationMsg := "Notification: Your contract in" + study.Title + " is due to expire soon"
 	return c.createCustomEmail(ctx, notificationMsg, emails, content)
 }
+
+func (c *Controller) SendTrainingExpiryNotification(ctx context.Context, emails []string, training types.UserTrainingRecord) error {
```
t-young31: why send to more than one person?

internal/types/training.go
```diff
@@ -21,3 +21,16 @@ type UserTrainingRecord struct {
 	// Relationships
 	User User `gorm:"foreignKey:UserID"`
 }
+
+func (t UserTrainingRecord) DaysUntilExpiry() int {
+	return int(time.Since(t.CompletedAt).Hours()/24) - 365
```
t-young31: magic numbers are bad. should reference the const https://github.com/ucl-arc-tre/portal/blob/e7a6c8bdf15474d3ea3c8cef9f14a646ac123fd7/internal/config/main.go#L28

internal/types/training.go
```diff
@@ -21,3 +21,16 @@ type UserTrainingRecord struct {
...
+}
+
+func (t UserTrainingRecord) CertificateExpiringWithin30Days() *UserTrainingRecord {
+	if t.DaysUntilExpiry() < 0 || t.DaysUntilExpiry() < 30 {
```
t-young31: first condition is redundant(?)

internal/controller/entra/custom_mail.go
acholyn: agree, would like an email not to be sent if it doesn't align with one of those day numbers (see pr description)

internal/controller/entra/custom_mail.go
acholyn: it won't, but the custom email takes an array so I think I just persisted that up when I might not need to?

internal/types/training.go
acholyn: wasn't entirely sure about that, ty 🙏

internal/controller/entra/custom_mail.go
t-young31: this doesn't do that

internal/types/training.go
acholyn: is it still worth having as a separate func tho?

internal/controller/entra/custom_mail.go
acholyn: suggestions for how we do that pls

internal/controller/entra/custom_mail.go
```diff
@@ -133,20 +133,39 @@ func newTemplatedEmailContent(params EmailTemplateParams) (*string, error) {
...
+		content += "in " + fmt.Sprintf("%d", *days) + " days. Please sign in to the Portal to upload a new contract."
 	}
 
-	notificationMsg := "Notification: Your contract in" + study.Title + " is due to expire soon"
-	return c.createCustomEmail(ctx, notificationMsg, emails, content)
+	subject := "Notification: Your contract in" + study.Title + " is due to expire soon"
+	return c.createCustomEmail(ctx, subject, emails, content)
+}
+
+func (c *Controller) SendTrainingExpiryNotification(ctx context.Context, email string, training types.UserTrainingRecord) error {
```
t-young31, line 156: I've changed the logic in `SendContractExpiryNotification` and `SendTrainingExpiryNotification` so they either send or error – there's no silent skip, which I think is nicer for the caller

cc: @acholyn

internal/config/main.go
```diff
@@ -185,3 +185,19 @@ func ProcessIdentity() string {
...
+	expiresAt := trainingRecord.CompletedAt.Add(TrainingValidity)
+	return int(time.Until(expiresAt).Hours() / 24)
+}
+
+func ShouldNotifyTrainingExpiry(trainingRecord types.UserTrainingRecord) bool {
+	if trainingRecord.CompletedAt.IsZero() { // unset value
+		return false
+	}
+	daysUntilExpiry := DaysUntilTrainingExpiry(trainingRecord)
+	if daysUntilExpiry < -90 { // don't notify very expired training
```
t-young31, line 199: thoughts on this welcome. sending daily email forever seems like a bad thing to do

acholyn: yeah, fair

t-young31 (approved):

Issue #257, opened 2025-09-02
Asset expiry


jhughes982:
## Requirements

- [ ] UI indication for expired assets (refer to the contract expiry UI indicator) - no repercussions
- [ ] optional expiry date field - update current form to remove `required` flag

jhughes982:
> how about: if the asset has a contract, it has an expiry inherited from the contract, otherwise optional asset expiry?

I seem to remember we decided against the above option during TI planning. It's also not in the requirements. Are we good to not implement this?

acholyn:
> > how about: if the asset has a contract, it has an expiry inherited from the contract, otherwise optional asset expiry?
> 
> I seem to remember we decided against the above option during TI planning. It's also not in the requirements. Are we good to not implement this?

yep, fine not to as we decided it would increase complexity and then if there are multiple contracts attached it would fall apart anyway

jhughes982:
Will the asset expiry warnings follow the [same logic](https://github.com/ucl-arc-tre/portal/blob/main/web/src/components/shared/exports.tsx#L39) used for contracts and training?

acholyn:
> Will the asset expiry warnings follow the [same logic](https://github.com/ucl-arc-tre/portal/blob/main/web/src/components/shared/exports.tsx#L39) used for contracts and training?

Don't think so; they don't need to be updated. So just if it's past the date of expiry, it should indicate it has expired

unless @finleybacon you strongly feel differently/I've misunderstood and assets do get updated?

jhughes982:
> Don't think so; they don't need to be updated. So just if it's past the date of expiry, it should indicate it has expired

I meant should the expiry urgency level be calculated using [this same logic](https://github.com/ucl-arc-tre/portal/blob/main/web/src/components/shared/exports.tsx#L39)? i.e. the `critical, high, medium, and low` values.

acholyn:
don't think it's strictly necessary but for consistency let's go with yes

PR #612, opened 2026-04-17
Asset expiry

jhughes982:
## Works towards #257 

- This PR adds the frontend and backend code for the assets expiry indicator,
- `expires_at` for assets is now optional,
- Added a reusable `ExpiryWarning` frontend component,
- The Assets tab on the study page now shows the attention indicator (!) for assets with an approaching or expired expiry date,
- Frontend and backend tests updated,
- Go migration code added (to be removed when deployed to prod),
- Drive-by: moved `contract_ids` from `AssetBase` to the `Asset` response type as it wasn't ever used on `AssetBase`

Screenshots:
Screenshot 2026-04-17 at 16 45 42

Screenshot 2026-04-17 at 16 46 00

### Checklist

### DIFF
--- a/api/web.yaml
+++ b/api/web.yaml
@@ -1356,7 +1356,6 @@ components:
         - accessed_by_third_parties
         - third_party_agreement
         - status
-        - contract_ids
       properties:
         title:
           type: string
@@ -1425,12 +1424,6 @@ components:
             - awaiting
             - destroyed
           description: Status of the asset
-        contract_ids:
-          type: array
-          description: List of contract IDs associated with the asset (empty array if none)
-          items:
-            type: string
-
     Asset:
       allOf:
         - $ref: "#/components/schemas/AssetBase"
@@ -1441,6 +1434,7 @@ components:
             - study_id
             - created_at
             - updated_at
+            - contract_ids
           properties:
             id:
               type: string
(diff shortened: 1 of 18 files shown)


api/web.yaml
```diff
@@ -1351,13 +1351,11 @@ components:
         - legal_basis
         - format
         - requires_contract
-        - expires_at
         - has_dspt
         - stored_outside_uk_eea
         - accessed_by_third_parties
         - third_party_agreement
         - status
-        - contract_ids
```
t-young31, line 1360: ~are these not used on asset update?~ no because you can't update an assets contracts 👌🏼

internal/graceful/db.go
```diff
@@ -146,6 +147,34 @@ func migrateContracts(db *gorm.DB) {
 	log.Info().Msg("Migrated contracts")
 }
 
+// Drop the NOT NULL constraint on assets.expires_at to make expiry optional
+func migrateAssetExpiresAt(db *gorm.DB) {
```
t-young31: does gorm not handle this automatically in automigrate?

jhughes982: Hmm, I read that automigrate behaves non-destructively and will [not drop constraints](https://github.com/go-gorm/gorm/issues/6101).

acholyn (approved):

internal/graceful/db.go
t-young31: mind trying it? I'm pretty sure it's fine to go not null -> null

jhughes982: Nice, it turns out automigrate does indeed handle it. Thanks for the heads up!

[removing unneeded manual migration code](https://github.com/ucl-arc-tre/portal/pull/612/commits/afca4b289118a16966dfc52d9935515a071f1d0b)

t-young31 (approved):
🚀

PR #615, opened 2026-04-20
init studies import

### LINKED ISSUE #448: Import service for Studies, Assets & Contracts e.g. internal/service/studies/import.go
- [x] ~Needs to take studies.json as input~, then create studies, assets, and contracts, linking assets and contracts to the correct parent study
- ~Needs to return caseref -> study uid mapping~
- [x] Dependent on #470
- [x] #576

t-young31:
## Resolves #448

Adds v0 study import functions. Likely more refinement needed but with a view to minimising the merge conflicts I'd like to get this merged in well ahead of actually doing the migration.

> [!NOTE]
> I have deliberately favoured duplication over reuse of backend code as there is necessarily less validation that can be done. These endpoints/functions will be deleted once the studies migration has happened

### Checklist


internal/config/main.go
```diff
@@ -22,7 +22,7 @@ const (
 
 	TimeFormat     = time.RFC3339
 	DateFormat     = "2006-01-02" // YYYY-MM-DD format for dates
-	MaxUploadBytes = 1e7          // 10 MB
+	MaxUploadBytes = 1e8          // 100 MB
```
acholyn, line 25: is this temp or are we just upping it

acholyn (approved):

internal/types/studies.go
```diff
@@ -115,15 +115,15 @@ type AssetLocation struct {
...
+	SignatoryUserId  *uuid.UUID `gorm:"type:uuid"`
 	Title            string
-	ThirdPartyName   string
+	ThirdPartyName   *string
 	OtherSignatories *string // free text: could be name(s), email(s), UPN(s), organisations
 	Status           string  // proposed, active, expired
-	StartDate        time.Time
-	ExpiryDate       time.Time
+	StartDate        *time.Time
+	ExpiryDate       *time.Time
```
jhughes982, line 126: Whats the consequence of having a null start and/or expiry date on a contract? Are they prompted later on to enter the info or are we ok leaving them blank? Maybe one for @finleybacon

jhughes982:
My time is limited today to do a full review, but I've had a scan through the code and it all looks good to me.

internal/types/studies.go
finleybacon: unfortunately it is not a required field in the sharepoint so it is null for many records. that's it i think - so no they won't be prompted I don't think, we have to accept that some of the old data is not of great quality but draw a line and move on otherwise we could spend whole year prompting users to update

jhughes982: Makes sense, thanks @finleybacon

internal/config/main.go
t-young31: I think we need to up it, given the objects currently in sharepoint

PR #625, opened 2026-04-29
Dockerfile and lint update

jhughes982:
- This PR provides a few config updates as we were getting errors since updating to Go 1.26

### Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,10 +7,14 @@ repos:
       - id: end-of-file-fixer
         exclude: openapi
 
-  - repo: https://github.com/golangci/golangci-lint
-    rev: v2.11.4
+  - repo: local
     hooks:
       - id: golangci-lint
+        name: golangci-lint
+        language: system
+        entry: golangci-lint run
+        types: [go]
+        pass_filenames: false
         args:
           - "--timeout=5m"
 
--- a/internal/controller/s3/main.go
+++ b/internal/controller/s3/main.go
@@ -29,7 +29,7 @@ type ClientInterface interface {
 
 type Controller struct {
 	client   ClientInterface
-	uploader *awsS3Manager.Uploader
+	uploader *awsS3Manager.Uploader //nolint:staticcheck
 }
 
 func New() *Controller {
@@ -56,14 +56,14 @@ func New() *Controller {
 	)
 	controller := Controller{
 		client:   client,
-		uploader: awsS3Manager.NewUploader(client),
+		uploader: awsS3Manager.NewUploader(client), //nolint:staticcheck
 	}
 	return &controller
 }
 
 func (c *Controller) StoreObject(ctx context.Context, metadata ObjectMetadata, obj types.S3Object) error {
 	log.Debug().Any("metadata", metadata).Msg("Uploading S3 object")
-	_, err := c.uploader.Upload(ctx, &awsS3.PutObjectInput{
+	_, err := c.uploader.Upload(ctx, &awsS3.PutObjectInput{ //nolint:staticcheck
 		Bucket: aws.String(config.S3BucketName()),
 		Key:    aws.String(metadata.Key()),
 		Body:   obj.Content,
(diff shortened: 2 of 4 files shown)


internal/controller/s3/main.go
```diff
@@ -29,7 +29,7 @@ type ClientInterface interface {
 
 type Controller struct {
 	client   ClientInterface
-	uploader *awsS3Manager.Uploader
+	uploader *awsS3Manager.Uploader //nolint:staticcheck
 }
```
jhughes982, line 33: With the new lint approach (see below comment thread), I had to add the `nolint:staticcheck` as it was flagging an error when trying to commit. e.g.:

```
internal/controller/s3/main.go:32:12: SA1019: awsS3Manager.Uploader is deprecated: superceded by feature/s3/transfermanager. See https://github.com/aws/aws-sdk-go-v2/discussions/3306 (staticcheck) uploader *awsS3Manager.Uploader
```

The fix for the actual lint error can be done as a separate PR.

.pre-commit-config.yaml
```diff
@@ -7,10 +7,14 @@ repos:
...
-    rev: v2.11.4
+  - repo: local
     hooks:
       - id: golangci-lint
+        name: golangci-lint
+        language: system
+        entry: golangci-lint run
+        types: [go]
+        pass_filenames: false
         args:
```
jhughes982, line 18: I was getting errors locally when trying to commit:

```
golangci-lint............................................................Failed
- hook id: golangci-lint
- exit code: 3

Error: can't load config: the Go language version (go1.25) used to build golangci-lint is lower than the targeted Go version (1.26.0)
The command is terminated due to an error: can't load config: the Go language version (go1.25) used to build golangci-lint is lower than the targeted Go version (1.26.0)
```

Even though I had Go version 1.26 installed locally. Apparently, the issue is that golangci-lint (v2.11.4) was built with Go version 1.25.

This new config tells pre-commit to use whatever golangci-lint is installed on the system instead. This uses the same local approach as e.g. `go-fmt`.

Happy to discuss this further if an alternative approach is preferred.

jhughes982:
We would also need to add ` go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest` to our CI actions to make this work.

What do you think @t-young31 ? Is there an alternative approach?

edit:
I added it for now so we can evaluate if we are happy with this approach or not

.pre-commit-config.yaml
t-young31: I'm not a fan of this as: (1) it requires another out-of band install of a thing & keeping it suitably up to date and (2) we don't benefit from it skipping commits that have no go changes

jhughes982: Is there a solution you'd recommend?

t-young31: ideally find how your `pre-commit` is building with the wrong go version, remove it, then `pre-commit install`

jhughes982: Ah, I assumed the issue wasn't isolated to my machine. Do you not get the same lint error locally when you make Go changes?

If it is isolated to my machine then yes I'll need to figure out what's caused it.

t-young31: nope – it's okay for me (after bump & `pre-commit install`)

jhughes982: Hmm, I've not ran in to this kind of issue in a long time. I've cleared all caches several times, reinstalled pre-commit and I'm still getting the error 😞

jhughes982: @acholyn I assume everything works ok on your end too? (try committing something after making a trivial change to a Go file).

t-young31: maybe setting

`language_version: 1.26.2` in the [pre-commit config](https://github.com/ucl-arc-tre/portal/blob/3c462d4a00f9cb33a4ebee9c5b8b790e186954c3/.pre-commit-config.yaml#L12) will force it?

jhughes982: That does work! But it doesn't address why the issue is happening only on my machine. Are you happy to add that as a config to work around my issue? If so, I'll open another small PR adding that config setting in.

jhughes982:
Closing in favour of #628

Issue #348, opened 2025-10-24
Glossary & definitions


acholyn:
We have definitions for some of the entities we've developed thus far, but as we include other services, we'll need to distinguish these so that users understand the structure we use.

Initially, since the only comparison would be to DSH, I think the following would be fine, 

<img width="762" height="323" alt="Image" src="https://github.com/user-attachments/assets/073c0102-4842-456f-b45c-e5cac00d1b04" />

but as we include more services we'll want a cleaner way of doing so

<img width="913" height="348" alt="Image" src="https://github.com/user-attachments/assets/ea02aa7c-dde0-418c-9326-7f5e00fc1526" />

It will also bear thinking about if we are completely transparent with users, eg with Myriad, if there's always a hidden study so that we can use the same data structures

Issue #517, opened 2026-02-25
CSS/UI conformance


acholyn:
## Description

We need to go through our current styling and component usage and determine:
- if we're not using an existing UIkit component when we could, why not and feed back the performance issues
- if there isn't a component available, what is it
- if we're using the component and overriding styling, why and
  - is this something fixed by the update
  - feed back so they can potentially expand the component

output as a word doc and share with Jason and Vincent

see also #461 

This will also be helpful to know where to look once we update the kit with the new branding

>[!TIP]
> check for where we use `!important`

acholyn:
related [doc](https://liveuclac.sharepoint.com/:w:/r/sites/RITSstaff/Shared%20Documents/ARC%20Services%20Portal/CSS%20conformance.docx?d=w4180d3c9399f44b780fec2e611c3fe30&csf=1&web=1&e=3KUFVy) to review separately

PR #661, opened 2026-05-19
Acholyn/UI tweaks

### LINKED ISSUE #495: ui consistency/tweaks
- [ ] remove assets title thing since study title and assets are already on the page
  - [ ] move callout into an info dropdown
  - [ ] add contracts info callout
- [ ] make study cards clickable, not just manage button
- [ ] more consistency between asset cards and contract cards (and project cards)
- [ ] add contract + asset summary to study card?
- [ ] fix flashing content on manage study page
- [ ] remove admin bullet points on study overview

acholyn:
### Checklist

- fixing bits broken by UI kit
- making things more consistent across assets + contracts + projects
- card component (wholly clickable)
- drive by: definitions exports for more consistency; related to #348 ?

see card component + new branding:
image

image

image
image
image


web/src/components/shared/entityDefinitions.tsx
```diff
@@ -0,0 +1,25 @@
+import InfoTooltip from "../ui/InfoTooltip";
+
+export const entityDefinitions = {
+  study: "Studies are a top level entity that can contain Assets and Contracts and own Projects",
```
acholyn, line 4: would like ideas for how to use definitions and the callout in one, maybe it just has to be separate components?

t-young31:
maybe for another PR: can we switch to new UCL styling for nav and top bar?

Screenshot 2026-05-26 at 09 15 28

***

also: remove all the rounded corners

t-young31 (commented):
I like the screenshots. will leave @jhughes982 to approve

acholyn:
> maybe for another PR: can we switch to new UCL styling for nav and top bar?
> 
> Screenshot 2026-05-26 at 09 15 28


yeah, I'm not entirely sure which component that is because I updated it to use `HeaderNew` as directed. can look into it for a future pr, maybe as part of #517 

> also: remove all the rounded corners

😭

web/src/styles/globals.css
```diff
@@ -41,6 +45,13 @@ body {
   -moz-osx-font-smoothing: grayscale;
 }
 
+button {
+  transition:
+    color 0.3s ease,
+    background-color 0.3s ease,
+    border-color 0.3s ease !important;
+}
```
jhughes982: Do we want all buttons to take on this behaviour?

web/src/components/ui/Card.tsx
jhughes982, line 1: The file is called Card but the component is called EntityCard, any reason for the distinction? Personally I'd just call it `Card` throughout.

web/src/components/ui/Card.tsx
```diff
@@ -0,0 +1,59 @@
...
+    footerContent,
+    manageUrl,
+    canModify = false,
+    isWarning,
+    deleteButton,
+  } = props;
+  const router = useRouter();
+  return (
+    <Box isCard key={key} warning={isWarning}>
+      <a href={manageUrl} data-cy="entity-card">
```
jhughes982: Two elements handle navigation here: `<a>` and `<Button>`. I don't think this is good for accessibility and may not be valid HTML. I'd consider a restructure here.

web/src/components/ui/Card.tsx
```diff
@@ -0,0 +1,59 @@
...
+        <div className={styles.header}>
+          <h2 className={styles.title}>{title}</h2>
+          {headerContent}
```
jhughes982: I don't think there is a corresponding `.title` css class in `Card.module.css` so this would render without styling.

web/src/components/studies/StudyCardsList.module.css
jhughes982, line 1: Theres some commented out code here that should be removed.

web/src/components/ui/Card.tsx
```diff
@@ -0,0 +1,59 @@
...
+    children,
+    footerContent,
+    manageUrl,
+    canModify = false,
+    isWarning,
+    deleteButton,
+  } = props;
+  const router = useRouter();
+  return (
+    <Box isCard key={key} warning={isWarning}>
```
jhughes982: Is the `key` prop here needed?

jhughes982: I would avoid implementing `Box` in this component, it feels like it should exist separately.

jhughes982 (commented):
Nice. A few comments to tweak some of the implementations.

web/src/styles/globals.css
acholyn: I couldn't think of where we didn't want a smooth transition with our buttons and UI kit apparently doesn't do it  built into the buttons

web/src/components/ui/Card.tsx
acholyn: touché, this was me getting ahead of myself https://github.com/ucl-arc-tre/portal/pull/661/commits/b1c416b31c2788e52476adc9c4d790b490685e6a

web/src/components/ui/Card.tsx
acholyn: I wanted the whole card to be clickable. Having looked into it again, it seems swapping the `button` for a `span` is acceptable or just rm the router push on click. Any thoughts?

web/src/components/ui/Card.tsx
jhughes982: How did you resolve this?

web/src/styles/globals.css
jhughes982: I would still use a custom class rather than apply it to all HTML button elements.

acholyn: have put it on our button component 
not sure how I forgot we had that

web/src/components/ui/Card.tsx
acholyn: https://github.com/ucl-arc-tre/portal/pull/661/commits/b8e9ce291ca9ffc53dea868920f2285476851415

web/src/components/ui/Card.tsx
jhughes982: Yes you could swap the button for any other element and just not attach a link to it, but at this point is there any need for the "View" button at all if the entire card is clickable? I think I'm ok with either approach (button vs entire clickable card) but maybe not both at the same time? Happy to hear other opinions though.

web/src/styles/form.css
jhughes982, line 1: Just call this file `form.css`?

web/src/pages/_app.tsx
```diff
@@ -4,17 +4,30 @@ import { AuthProvider } from "@/hooks/useAuth";
...
+
+    loadTheme();
+  }, []);
+
   return (
     <AuthProvider>
       <MetaHead title="ARC Services Portal | UCL" description="ARC Services Portal" />
 
-      <ThemeContextProvider>
+      <ThemeContextProvider initialTheme={lightTheme ? lightTheme : undefined}>
```
jhughes982: I think this expression is redundant, it could just simply be `lightTheme`.

web/src/pages/_app.tsx
```diff
@@ -4,17 +4,30 @@ import { AuthProvider } from "@/hooks/useAuth";
...
+  const [lightTheme, setLightTheme] = useState<ThemeType | undefined>(undefined);
+
+  useEffect(() => {
+    async function loadTheme() {
+      const { lightTheme } = await import("uikit-react-public/dist/");
+      setLightTheme(lightTheme);
+    }
+
+    loadTheme();
+  }, []);
```
jhughes982, line 27: Is this the recommended way to import the theme? Does a simple import without having to use `useEffect` not work or does it throw errors (I seem to remember some ongoing import issues)?

acholyn: it's not the favoured way of importing the theme but there were a lot of issues with the favoured way (namely that it didn't work) so Jason and I had a back and forth on how to get it to work, this was the only thing I tried that ended up working. They have more work to do with the theme anyway but are aware of our issue of not being able to import it normally. Hopefully we'll work something out

jhughes982: ok, can you add a brief contextual comment here, and possible link to an issue ticket if there is one?

web/src/components/ui/Card.tsx
acholyn: I guess my concern would be when there are other button options (eg delete) then it'll be less obvious to view the item if there's not also a view/manage button? Could update the conditional so if there is other footer content then show the manage/view button?
imo limiting the view/manage to only the button for it feels outdated/annoying

might be best to leave it and see what DX thinks when they review our UX?

jhughes982: I see your point. The only other thing would be that the HTML structure is now slightly convoluted. For instance, we have a `Button` component with a css class of `actions`, but the button itself doesn't do anything as the action (i.e. redirect) is controlled by the parent `<a>` element.

I'll leave the final decision to you.

web/src/pages/_app.tsx
acholyn: context: dynamic imports work for components. The theme can't be imported as a variable for the same reason the components can't be imported (unclear) but the dynamic method doesn't work for the theme (not a component). Workaround is to fetch and assign the theme within a component and pass it to children
issue: https://github.com/ucl-isd/uikit-react/issues/311

jhughes982: I meant add a comment in the code for contextual reference 😅

web/src/components/ui/Card.tsx
acholyn: have gone with whole card is link if it's only to view, if delete then will show buttons and whole card not clickable. if DX team hates it, will update in due course but HTML is happy and should be good accessibility-wise

web/src/components/ui/Card.tsx
```diff
@@ -0,0 +1,47 @@
+import { useRouter } from "next/dist/client/components/navigation";
```
jhughes982: The correct import should be: `import { useRouter } from "next/router";`

But you shouldn't need to use `useRouter` here as the `Link` component from `next/link` is the cleaner solution.

web/src/components/projects/ProjectCardsList.tsx
```diff
@@ -80,39 +79,30 @@ export default function ProjectCardsList(props: Props) {
...
-                  }
-                  size="small"
-                >
-                  {`Manage Project ${isOpsStaff ? "Approval" : ""}`}
-                </Button>
-                {!isOpsStaff && (
+            <Card
+              key={project.id}
+              title={project.name}
+              manageUrl={`/projects/manage?projectId=${project.id}`}
```
jhughes982: This URL is missing the `environment` param which means the navigation won't work.

web/src/components/contracts/ContractObjectCard.tsx
```diff
@@ -88,61 +89,55 @@ export default function ContractObjectCard(props: ContractObjectCardProps) {
...
+              data-cy="contract-object-delete-button"
+            >
+              Delete
+            </Button>
+          )}
+        </div>
+      }
+    >
+      <>
+        <span className={styles.createdAt}>Created at: {createdAt}</span>
```
jhughes982: The `createdAt` style doesn't exist in `ContractObjectCard.module.css` which means this will return `undefined`.

web/src/components/contracts/ContractManagement.tsx
```diff
@@ -35,18 +38,33 @@ export default function ContractManagement(props: ContractManagementProps) {
...
+            <span className={styles["contracts-count-badge"]}>{contracts.length}</span>
+          </div>
         </>
       ) : (
         <div className={styles.header}>
           <h3>Contracts</h3>
+          <div className={styles["contracts-summary"]}>
+            <span>Number of associated Contracts:</span>
+            <span className={styles["contracts-count-badge"]}>{contracts.length}</span>
+          </div>
```
jhughes982: The conditional renders the exact same JSX: ` <div className={styles["contracts-summary"]}>`. Is that intentional?

web/src/components/ui/Box.tsx
```diff
@@ -1,5 +1,5 @@
 import styles from "./Box.module.css";
 
-export default function Box({ children }: { children: React.ReactNode }) {
+export default function Box({ children }: { children: React.ReactNode; warning?: boolean }) {
   return <div className={styles.container}>{children}</div>;
```
jhughes982: The `warning` prop is never used here.

web/src/components/ui/Card.tsx
```diff
@@ -0,0 +1,47 @@
...
+              Manage
+            </Button>
+
+            {deleteButton}
+          </div>
+        )}
+      </div>
+    </article>
+  );
+  return manageUrl && !deleteButton ? <a href={manageUrl}>{CardInner}</a> : CardInner;
```
jhughes982: You should be able to use the `Link` component from nextjs here

web/src/components/ui/Card.tsx
```diff
@@ -0,0 +1,47 @@
...
+              size="small"
+              className={styles.manage}
+              data-cy="manage-button"
+              onClick={() => router.push(manageUrl)}
+            >
+              Manage
+            </Button>
+
+            {deleteButton}
+          </div>
```
jhughes982: Personally I'm not a fan of implementing a card this way, there are a few issues from my perspective:
1. We now have inconsistent UX where users sometimes click the entire card to navigate to the manage page and sometimes they don't (i.e. for projects). 
2. Card components should ideally be pure UI components with props passed in. This implementation forces you to read more logic to understand how the card works in terms of its linking behaviour and delete button rendering. A cleaner solution might be to just pass the delete button into the `footerContent`.

Overall, I don't see a massive benefit of making the entire card clickable if it adds code complexity and UX inconsistency. Happy to be overruled on this one though if I'm in the minority. Alternatively, there may be a way to implement the best of both worlds, e.g. perhaps we remove the "Manage" button entirely so the user always clicks the entire card to navigate?

Happy to hear more thoughts @t-young31 @finleybacon

web/src/components/contracts/ContractManagement.tsx
acholyn: I think the difference is in the header shown above the summary
Contracts vs Contract Management + button + callout

jhughes982: Is there no way to remove the duplicated JSX? Either as a variable or moving the HTML structure around?

web/src/components/ui/Card.tsx
t-young31: agree. happy with either option as long as we're consistent

acholyn: >we remove the "Manage" button entirely so the user always clicks the entire card to navigate

went with this
I'm not entirely sure why we have delete on the project card and not asset or study cards (it makes sense with the contracts), so I'm happy to discuss that and see if we all think it makes sense (out of scope of this PR)

acholyn: ^ agreed we'll rm the delete from the project card in another PR

web/src/components/projects/ProjectCardsList.tsx
```diff
@@ -73,39 +72,30 @@ export default function ProjectCardsList(props: Props) {
...
-                  size="small"
-                >
-                  {`Manage Project ${isOpsStaff ? "Approval" : ""}`}
-                </Button>
-                {!isOpsStaff && project.status === "incomplete" && (
+            <Card
+              key={project.id}
+              title={project.name}
+              manageUrl={`/projects/manage?projectId=${project.id}&environment=${project.environment_name}`}
+              headerContent={<StatusBadge status={project.status} type="project" />}
```
jhughes982: Does this need to take an `environment` prop?

web/src/components/assets/AssetCard.tsx
```diff
@@ -57,19 +57,49 @@ export default function AssetCard(props: AssetCardProps) {
...
+              </small>
+              <Button
+                onClick={() => {
+                  router.push({ query: { ...router.query, tab: "contracts" } }, undefined, { shallow: true });
+                  window.scrollTo({ top: 0, behavior: "smooth" });
+                }}
+                variant="secondary"
+                size="small"
+              >
+                Contracts
```
jhughes982: This now has the same linking issue as previously discussed. Because the card itself is now wrapped with an `<a>` the linking behaviour on the `<Button>` causes a conflict. Have you tested this locally to see what happens? Does the button take you to the correct url?

jhughes982 (commented):
A few more observations.

Also, there is a `console.log(status, getStatusClassName(status))` in `StatusBadge.tsx` that you could remove as a drive-by.

web/src/components/projects/ProjectCardsList.tsx
acholyn: don't believe so; it's just part of the url

web/src/components/assets/AssetCard.tsx
acholyn: ah, missed this. does not, will rm

web/src/components/projects/ProjectCardsList.tsx
jhughes982: `StatusBadge` does indeed require the `environment` prop. Otherwise the project badge tooltip text will default to `Status information not available`.

acholyn: ahh, I misunderstood which part it was for. have added https://github.com/ucl-arc-tre/portal/pull/661/commits/c9ece5f58fc8629edb158b4d5e2d7ca975385c4d

jhughes982 (approved):
Looks good!

Issue #605, opened 2026-04-14
IG User Testing + Feedback


acholyn:
For doing migration we need to test the user experience for the IG team.

Would like to do a semi unstructured engagement such that we can check that the experience is intuitive and the process is easy to do.

Should liaise with the UX team to help guide the feedback 
Asks from DX:
- objectives/purpose of portal etc. requirements, ux to create scenarios, access to portal, 2 weeks notice
- demo
- 10 people, give them a month or so notice

---

### Functionality Requirements
- [ ] view Study and connected items
- [ ] able to approve Study or leave feedback
- [ ] confident in the information shown

acholyn:
will see if IG team available to do this next week
update: session booked for Thurs June 4

acholyn:
another (final) round scheduled for aug 18 and should then be able to return to TRE migration

PR #682, opened 2026-06-05
misc refactoring/fix from UAT

t-young31:
Small bits of refactoring from #605

### Checklist

### DIFF
--- a/internal/tasks/expiry_checks.go
+++ b/internal/tasks/expiry_checks.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"time"
 
+	"github.com/rs/zerolog/log"
 	"github.com/ucl-arc-tre/portal/internal/config"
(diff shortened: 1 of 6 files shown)


internal/tasks/expiry_checks.go
```diff
@@ -121,6 +122,7 @@ func (m *Manager) checkStudySignoffExpiry() error {
 			continue
 		}
 
+		log.Debug().Str("study", study.Title).Any("owner", study.Owner.Username).Msg("Notifying study signoff")
```
acholyn, line 125: are these temp or are we gonna keep them for a while?

t-young31: I think useful at `debug` level, so I vote keep. at some point we turn debug off in prod

acholyn (approved):
👍

PR #706, opened 2026-06-10
putting risk score on asset card

acholyn:
### Checklist

image


t-young31 (approved):
nice. we might want to colour code the value so it's more meaningful, but LGTM for now

acholyn:
> nice. we might want to colour code the value so it's more meaningful, but LGTM for now

yeah, but we don't know what the thresholds are atm anyway, pending disucssion afaik

PR #716, opened 2026-06-11
adding guidance on dpo

### LINKED ISSUE #700: 1.1.10 DPO registration question needs guidance – High priority

acholyn:
## Resolves #700

does not show if yes

image

### DIFF
--- a/web/src/components/studies/study-form/StudyFormStep3.tsx
+++ b/web/src/components/studies/study-form/StudyFormStep3.tsx
@@ -27,20 +27,7 @@ export default function StudyFormStep3({ control, errors, controllerValue }: Stu
         />
       </div>
       <div className={sharedStyles["option-field"]} data-cy="isDataProtectionOfficeRegistered">
-        <div>
-          The research is already registered with the UCL Data Protection Office
-          <HelperText>
-            <div className={styles["ucl-guidance"]}>
-              <p>
-                All research involving personal data must be registered before any processing begins. See{" "}
-                <a href="https://www.ucl.ac.uk/data-protection/guidance-staff-students-and-researchers/research/research-registration-guidance">
-                  guidance
-                </a>
-                .{" "}
-              </p>
-            </div>
-          </HelperText>
-        </div>
+        <div>The research is already registered with the UCL Data Protection Office</div>
         <Controller
           name="isDataProtectionOfficeRegistered"
           control={control}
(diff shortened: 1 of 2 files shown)


acholyn:
or if it should be regardless of pos selection:
image

which I don't love but might make more sense

web/src/components/studies/study-form/StudyFormStep3.tsx
```diff
@@ -38,73 +38,84 @@ export default function StudyFormStep3({ control, errors, controllerValue }: Stu
...
+        <>
+          <div className={styles["ucl-guidance"]}>
+            <p>
+              All research involving personal data must be registered before any processing begins. See{" "}
+              <a href="https://www.ucl.ac.uk/data-protection/guidance-staff-students-and-researchers/research/research-registration-guidance">
+                guidance
+              </a>
+              .{" "}
+            </p>
+          </div>
```
t-young31: I think this should hide if yes is registered

web/src/components/studies/study-form/StudyFormStep3.tsx
```diff
@@ -38,73 +38,84 @@ export default function StudyFormStep3({ control, errors, controllerValue }: Stu
...
-                    id="dataProtectionPrefix"
-                    readOnly={isUcl}
-                    placeholder={isUcl ? "" : "Registry ID eg ZX1234"}
-                    className={isUcl ? sharedStyles.readonly : ""}
-                  />
-                );
-              }}
-            />
+        <>
+          <div className={styles["ucl-guidance"]}>
```
t-young31: should be a ~warning~ alert not something custom 😒

acholyn: i just grabbed the styling we use for guidance on the asset form 🤷

web/src/components/studies/study-form/StudyFormStep3.tsx
acholyn: feels unnecessary, will just leave as part of the q; conditional to hide feels bad

web/src/components/studies/study-form/StudyFormStep3.tsx
acholyn: image



this is as alert type=warning
no me gusta 👎

t-young31 (approved):
🚢

PR #736, opened 2026-06-15
fixing msword parsing

acholyn:
### Checklist

I'm not convinced this is the best way of doing this but was lacking ideas


internal/validation/contract.go
```diff
@@ -16,6 +16,7 @@ var (
 		types.MimeTypeDocx,
 		types.MimeTypeJpeg,
 		types.MimeTypePNG,
+		types.MimeTypeOctetStream,
```
t-young31: I'd prefer not to allow this uniformly, rather only if it's a `.doc` file

internal/handler/web/contracts.go
```diff
@@ -82,6 +85,16 @@ func (h *Handler) PostStudiesStudyIdContractsContractIdObjects(ctx *gin.Context,
 	} else if !validation.IsValidContractMimeType(mimeType) {
 		setError(ctx, types.NewErrInvalidObjectF("mime type was [%v] not valid", mimeType), "Invalid MIME type")
 		return
+	} else if strings.ToLower(filepath.Ext(fileHeader.Filename)) == ".doc" &&
+		(mimeType != types.MimeTypeOctetStream &&
+			mimeType != types.MimeTypeDoc) {
+
+		setError(ctx, types.NewErrInvalidObjectF("mime type was [%v], did not match valid type for doc", mimeType), "Invalid MIME type for filetype")
+	} else if strings.ToLower(filepath.Ext(fileHeader.Filename)) == ".docx" &&
+		(mimeType != types.MimeTypeZip &&
```
t-young31: this is not allowed?

t-young31 (commented):
can we make some sample .doc and .docx and upload them in an e2e to check that this works?

internal/validation/contract.go
acholyn: how are you envisioning I do that

internal/handler/web/contracts.go
acholyn: my thinking was if it's .docx and NOT zip or docx type then should throw error

t-young31: I think it should default disallow zip and octect/stream, then whitelist only `.doc`, `.docx` extensions. happy to push to this branch if you'd like

acholyn: yes pls

t-young31: [eb61cec](https://github.com/ucl-arc-tre/portal/pull/736/commits/eb61cec661ee6525d337bff28868c4409b283f02)

t-young31 (approved):

PR #742, opened 2026-06-18
adding blurb to say access will be revoked on env level

acholyn:

### DIFF
--- a/internal/controller/entra/custom_mail.go
+++ b/internal/controller/entra/custom_mail.go
@@ -176,7 +176,7 @@ func (c *Controller) SendTrainingExpiryNotification(ctx context.Context, email s
 		content += "in " + fmt.Sprintf("%d", days) + " days. Please sign in to the Portal to upload a new certificate."
 	}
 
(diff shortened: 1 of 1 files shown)


internal/controller/entra/custom_mail.go
```diff
@@ -176,6 +176,8 @@ func (c *Controller) SendTrainingExpiryNotification(ctx context.Context, email s
 		content += "in " + fmt.Sprintf("%d", days) + " days. Please sign in to the Portal to upload a new certificate."
 	}
 
+	content += "<br><br>Please note that without valid training, your access to any environments will be revoked."
```
t-young31:
```suggestion
	content += "<br><br>Please note that without valid training, your access to any environments may be revoked."
```

acholyn: >agree with TRE and DSH that access will be revoked on day of expiry

t-young31: also these `<br>`s don't render

t-young31:
can you replace the other `<br>` pls

t-young31 (approved):

PR #744, opened 2026-06-18
date format consistency

### LINKED ISSUE #611: Use of datetime formats on the frontend
- We currently have two exposed methods for formatting dates on the frontend: `convertRFC3339ToDDMMYYYY` and `formatDate`,
- Are both needed? Do we want to display datetimes differently across different areas of the portal?
- In addition, some areas of the portal display expiry dates (e.g. contracts) along with a local datetime. Does it make sense to display the time segment here? Are expiry dates sensitive to exact hours/minutes?

## Requirements
- [x] Clarify if we need multiple datetime functions for the frontend
- [x] Clarify if hours/minutes are approrpiate to display for expiry 
  - [ ] If hours/minutes are not needed for expiry, then remove it

## Acceptance Criteria
- [ ] keep the convert... function unless the places where it's been used no longer need it because maybe we've updated the underlying date format?
- [ ] rename formatDate so it's clear that it's human readable eg. `humanReadableOfDDMMYY` (UCL has people from all over, I'd rather be sure we all know 01/04 is April 1 and not maybe Jan 4)
- [ ] remove the other usage of formatDate and just go with the one with the time
- [ ] we don't really need the time for expiry, remove it

acholyn:
### Checklist

image
image


web/src/components/shared/exports.tsx
```diff
@@ -14,14 +14,12 @@ export function convertRFC3339ToDDMMYYYY(dateString: string) {
   return `${day}/${month}/${year}`;
 }
 
-export function formatDate(dateString: string) {
+export function humanReadableOfDDMMYYYY(dateString: string) {
```
t-young31: I'm weakly in favour of `formatDate`

web/package.json
```diff
@@ -21,7 +21,7 @@
     "react-is": "^19.2.7",
     "react-markdown": "^10.1.0",
     "recharts": "^3.8.1",
-    "uikit-react-public": "^0.25.5"
+    "uikit-react-public": "^0.26.0"
```
t-young31: what does this do?

acholyn: updates the kit, has the newer theme thing, and pagination
theme thing doesn't work and haven't tried pagination yet

web/src/components/shared/exports.tsx
acholyn: I want a more obvious name to distinguish it from `convertRFC3399...` and formatDate feels like it could apply to the convert stuff
open to other names tho

t-young31: I'd prefer `formatDate` and `formatTime`, with a comment linking back to the API definition where the formats are defined

acholyn: which would be which?

t-young31: the one with a time should be `formatTime`...

acholyn: neither has time...
or am I going mad

acholyn: Image??

t-young31: rfc3339 includes time

t-young31 (approved):
🚢

PR #750, opened 2026-06-24
add users lookup

### LINKED ISSUE #391: Refactor user validity lookup (for studies and projects)
- At the moment the study form (and soon the projects form) look up the validity of a user at the point of submission. So the error gets returned after the lookup has completed.
- A preferred approach would be to run the user validation check dynamically at the point of adding them to the form. So in this case the user would type in the username and presumably click an `add` button where it would then run the validity check?

Tagging @t-young31 and @acholyn for thoughts

acholyn:
### Checklist
image
image

TODO:
- confirm we're doing away with 1 IAA limit


acholyn:
do we like the white bg or the blue bg? or should it be purple now 😆 🤔 
image
image

@finleybacon @nikorobins @t-young31 @jhughes982 thoughts?

t-young31:
no preference

nikorobins:
Hi @acholyn we don't have the blue color in the new brand but you can check our palette here https://verbose-adventure-l1ov64q.pages.github.io 

The issue here is more using an alert (success or info) for a purpose that user do not expect.

acholyn:
> Hi @acholyn we don't have the blue color in the new brand but you can check our palette here https://verbose-adventure-l1ov64q.pages.github.io
> 
> The issue here is more using an alert (success or info) for a purpose that user do not expect.

do you have a component that looks similar?

internal/controller/entra/custom_mail.go
```diff
@@ -108,15 +108,24 @@ func (c *Controller) sendCustomEmail(ctx context.Context, subject string, emails
...
 
-	content := ""
+	content := "You have been invited to join the"
+	if studyName != nil {
+		if projectName != nil {
+			content += " project '" + *projectName + "' in the study '" + *studyName + "'"
+		} else {
+			content += " study '" + *studyName + "'"
+		}
+	}
```
t-young31: if `studyName` is nil this is not a good sentence

internal/controller/entra/custom_mail.go
```diff
@@ -108,15 +108,24 @@ func (c *Controller) sendCustomEmail(ctx context.Context, subject string, emails
...
+
 	if sponsor.ChosenName != "" {
-		content = "You have been invited to join the UCL ARC Services Portal by " + string(sponsor.ChosenName)
+		content += " by " + string(sponsor.ChosenName)
 	} else {
-		content = "You have been invited to join the UCL ARC Services Portal by " + string(sponsor.Username)
+		content += " by " + string(sponsor.Username)
 	}
+	content += "\nYou can view this on the UCL ARC Services Portal"
```
t-young31: "this" = ?

go.mod
t-young31, line 1: revert pls

api/web.yaml
```diff
@@ -208,6 +208,12 @@ paths:
           description: user details to lookup by in entra. This can be valid within the user principal name, email, given name or display name eg. "tom", "hughes", "ccaeaea", "laura@example"
           schema:
             type: string
+        - in: query
+          name: is_approved_researcher
```
t-young31: the backend for this filter is not implemented...

t-young31:
😕

Screenshot 2026-07-08 at 10 51 17

web/src/components/shared/UserLookup.tsx
```diff
@@ -0,0 +1,282 @@
...
+  usernames: StudyFormData["additionalStudyAdminUsernames"] | ProjectFormData["members"];
+  appendUsername: (value: string) => void;
+  removeUsername: (username: string) => void;
+  roleControl?: ReactElement | ((user: UserData) => ReactElement | null);
+  studyName?: string;
+  projectName?: string;
+  limit?: number;
+  filterIAO?: string;
+};
+export default function UserLookup(props: UserLookupProps) {
```
t-young31, line 66: this re-renders when roles change – I don't love

t-young31: resolved by [5251bb8](https://github.com/ucl-arc-tre/portal/pull/750/commits/5251bb8b39d6b76bd285042771c22c83315afdfe)

t-young31:
> confirm we're doing away with 1 IAA limit

I'm not aware that we are. has anything changed?

go.mod
t-young31: resolved by 118180a6

finleybacon:
Like the lookup, very cool 😎 however when logged in as portal-fb-staff, if add my portal-fb-admin user as an IAA, and then go to view that Study as portal-fb-admin, the IAA agreement doesn't come up, so I can't sign it. maybe this is dev environment related but I can't think what would be causing it.

also from the issue:

> if person not found, show invite button

this isn't implemented right?

t-young31:
> portal-fb-admin user as an IAA

think admin gets a different view. do you have another account that you can add as an IAA?

t-young31 (approved):

t-young31:
download

PR #764, opened 2026-07-07
tre project versioning

### LINKED ISSUE #762: Add TRE project versioning
**As a** project owner
**I want** to know if my TRE project is awaiting an update
**So that** I know who has access

## Options

a. Use `updated_at`
b. Version number
c. Hash of some sort

## Acceptance criteria

- [ ] Auto incrementing version number
- [ ] TRE API `POST /projects/{name}` to update deployed version

t-young31:
## Resolves #762

- Changes `/projects/{projectName}/status` -> `/projects/{projectName}` for multiple field updates in the TRE API
- Adds `RequestedVersionUpdatedAt` and `DeployedVersionUpdatedAt` to track when a new version of the project was requested and the version that is deployed

### TODO

- [x] e2e test

### Checklist


internal/types/projects.go
```diff
@@ -29,6 +30,15 @@ type ProjectTRE struct {
 	AirlockWhitelist              ProjectTREWhitelist `gorm:"serializer:json"`
 	Status                        ProjectTREStatus    `gorm:"not null;default:'incomplete'"`
 
+	// Version of the project which has been requested
+	RequestedVersionUpdatedAt *time.Time
```
t-young31, line 34: thought we could use the `ModelAuditable.UpdatedAt` value, but that changes once on set of `DeployedVersionUpdatedAt` from the TRE API, so the UpdatedAt times are never in sync.

other suggestions welcome

internal/service/projects/main.go
milanmlft: Dead?

internal/service/projects/main.go
```diff
@@ -525,6 +530,38 @@ func (s *Service) CreateTREVMImage(data treopenapi.VMImage) error {
...
 }
 
+func (s *Service) UpdateProjectTREDeployed(projectName string, data treopenapi.ProjectUpdate) error {
+	if !data.Status.Valid() {
+		return types.NewErrClientInvalidObjectF("invalid status")
+	}
+	status := types.ProjectTREStatus(data.Status)
+	if status != types.ProjectTREStatusDeployed && status != types.ProjectTREStatusDeleted {
+		return types.NewErrClientInvalidObjectF("status can only be deployed or deleted, was [%s]", status)
+	}
```
milanmlft, line 514: Isn't this covered already by `!data.Status.Valid()` above?

internal/service/projects/main.go
```diff
@@ -525,6 +530,38 @@ func (s *Service) CreateTREVMImage(data treopenapi.VMImage) error {
 	return types.NewErrFromGorm(result.Error, "failed to create TRE project vm image")
 }
 
+func (s *Service) UpdateProjectTREDeployed(projectName string, data treopenapi.ProjectUpdate) error {
+	if !data.Status.Valid() {
+		return types.NewErrClientInvalidObjectF("invalid status")
```
milanmlft, line 509:
```suggestion
		return types.NewErrClientInvalidObjectF("status can only be deployed or deleted, was [%s]", data.Status)
```

and remove the check below?

internal/types/projects.go
```diff
@@ -29,6 +30,15 @@ type ProjectTRE struct {
...
+
+	// Version of the project that has been deployedUses the value of ModelAuditable.UpdatedAt. i.e.
+	// nil => project has not been deployed
+	// ModelAuditable.UpdatedAt == DeployedVersionUpdatedAt =>  project deploy is up to date
+	// ModelAuditable.UpdatedAt.After(DeployedVersionUpdatedAt) => project is awaiting new version being deployed
```
milanmlft:
```suggestion
	// Version of the project that has been deployed. i.e.
	// nil => project has not been deployed
	// RequestedVersionUpdatedAt == DeployedVersionUpdatedAt =>  project deploy is up to date
	// RequestedVersionUpdatedAt.After(DeployedVersionUpdatedAt) => project is awaiting new version being deployed
```

stale comment?

api/web.yaml
```diff
@@ -1970,6 +1970,7 @@ components:
         - pending-approval
         - pending-creation
         - deployed
+        - pending-update
```
milanmlft: Add a case for `pending-update` to https://github.com/ucl-arc-tre/portal/blob/6f151dadbcf9ed4fa743bfc6a339338201fba335/web/src/components/ui/StatusBadge.tsx?plain=1#L32 too?

internal/handler/tre/main.go
```diff
@@ -54,23 +54,13 @@ func (h *Handler) GetUserStatus(ctx *gin.Context, params openapi.GetUserStatusPa
 	ctx.JSON(http.StatusOK, userStatus)
 }
 
-func (h *Handler) PostProjectsProjectNameStatus(ctx *gin.Context, projectName string) {
-	data := openapi.ProjectStatusUpdate{}
+func (h *Handler) PostProjectsProjectName(ctx *gin.Context, projectName string) {
+	data := openapi.ProjectUpdate{}
 	if err := ctx.ShouldBindBodyWithJSON(&data); err != nil {
 		setError(ctx, types.NewErrInvalidObject("failed to bind ProjectStatusUpdate"))
```
milanmlft:
```suggestion
		setError(ctx, types.NewErrInvalidObject("failed to bind ProjectUpdate"))
```

internal/service/projects/main.go
```diff
@@ -525,6 +530,38 @@ func (s *Service) CreateTREVMImage(data treopenapi.VMImage) error {
...
+	}
+	status := types.ProjectTREStatus(data.Status)
+	if status != types.ProjectTREStatusDeployed && status != types.ProjectTREStatusDeleted {
+		return types.NewErrClientInvalidObjectF("status can only be deployed or deleted, was [%s]", status)
+	}
+	deployedVersionUpdatedAt, err := time.Parse(config.TimeFormat, data.DeployedVersionUpdatedAt)
+	if err != nil {
+		return types.NewErrInvalidObject(err)
```
milanmlft: Should this not use `types.NewErrClientInvalidObjectF` too? Wrong time format should be fixable by the caller I think

internal/types/projects.go
milanmlft: Maybe silly question but is there ever a case when a TRE project is already deployed, `PUT /projects/tre/{projectId}` is called and the project status *shouldn't* be set to `pending-update`?
Because if not we could just skip these `UpdatedAt` time checks and just do something like

```go
if projectTRE.Status == types.ProjectTREStatusDeployed {
	projectTRE.Status = types.ProjectTREStatusPendingUpdate
}

```
in https://github.com/ucl-arc-tre/portal/blob/6f151dadbcf9ed4fa743bfc6a339338201fba335/internal/service/projects/main.go?plain=1#L397?

internal/types/projects.go
milanmlft, line 75: add `ProjectTREStatusPendingUpdate` here too?

milanmlft (commented):
Looking nice! Couple of comments and a potentially silly suggestion

internal/types/projects.go
t-young31: yes, when the deployed and requested project state is the same

milanmlft: hmm, but https://github.com/ucl-arc-tre/portal/blob/6f151dadbcf9ed4fa743bfc6a339338201fba335/internal/service/projects/main.go#L411 sets `RequestedVersionUpdatedAt` to current time on every `UpdateProjectTRE` call, so it'll always be newer than `DeployedVersionUpdatedAt`, no? I.e. `UpdateProjectTRE` will always result in `pending-udpate` status

milanmlft (commented):
Also, can we have some test coverage for `UpdateProjectTREDeployed`? :)

internal/types/projects.go
t-young31: yeah, so I was looking to _not_ store the `pending-update` state in the database, as it's derivable from the requested and deployed versions. requested after deployed => pending-update, so no need to store the duplicate bit of state (and have it potentially become inconsistent). if there's a good reason to have it though then I'm happy to be convinced

milanmlft: Ok I think it's fine then, can't think of a better way in that case

t-young31: could lower the confusion by keeping the frontend and backend statuses in sync, and just having a `pending_deployment_update` flag. might try it

internal/service/projects/main.go
t-young31: dead - merci

internal/service/projects/main.go
t-young31: nah. this is checking that the internal (db) enum and API enum haven't drifted

internal/service/projects/main.go
t-young31: I don't like coupling this comment to the API spec. it will go stale

api/web.yaml
t-young31: [now](https://github.com/ucl-arc-tre/portal/pull/764#discussion_r3543909004) n/a

internal/service/projects/main.go
t-young31: agree. should probably do this everywhere in the codebase at some point

internal/types/projects.go
t-young31: [now](https://github.com/ucl-arc-tre/portal/pull/764#discussion_r3543909004) n/a

LakshithadeSilva (approved):

Issue #756, opened 2026-07-01
Add `GET /projects` to TRE API


t-young31:
**As** tre ops staff
**I want** to be able to pull the projects configuration from the portal
**So that** TRE project CRUD can happen

### Acceptance criteria

- [ ] API updated https://github.com/ucl-arc-tre/portal/blob/cada4734ec6684b3e33fc716674a4ce933ed0608/api/tre.yaml#L9
- [ ] New TRE API handler method added https://github.com/ucl-arc-tre/portal/blob/cada4734ec6684b3e33fc716674a4ce933ed0608/internal/handler/tre/main.go#L26
- [ ] Response returns all TRE projects with all the roles etc. joined in

Issue #662, opened 2026-05-19
Refactor /internal/rbac package to use an instance-based design


yhong123:
The current `/internal/rbac` package relies on a singleton pattern with a global enforcer and database connection. This design does not support running integration tests in parallel.

As discussed with Tom, the package should be refactored to use instance-based components instead of global state, enabling isolated test environments and improved test concurrency.

PR #766, opened 2026-07-07
init tre project import

### LINKED ISSUE #765: Add TRE project import
**As a** TRE ops member
**I want** to be able to import current projects into the portal
**So that** they can be managed by their owners

## Acceptance criteria

- [ ] RBAC assigned
- [ ] Study links retained

t-young31:
## Resolves #765 

- Likely needs tweaking when we try it for real, and with #756. Thoughts appreciated
- Import code is https://github.com/t-young31/portal-project-import (pls ask for access if you want it)

### Checklist


internal/service/projects/main.go
```diff
@@ -562,6 +577,130 @@ func (s *Service) UpdateProjectTREDeployed(projectName string, data treopenapi.P
...
+			tx.Rollback()
+			return types.NewErrFromGorm(err, "failed to create ProjectTREUserConfig")
+		}
+	}
+
+	if _, err := rbac.AddProjectTreOwnerRole(project.StudyID, project.ID); err != nil {
+		tx.Rollback()
+		return err
+	}
+	return types.NewErrFromGorm(tx.Commit().Error, "failed to import tre project")
```
milanmlft, line 674: Is there a risk here that if `tx.Commit()` fails, the RBAC rules persist and so a user owns a project that doesn't exist in the portal?

If so, could evaluate the `tx.Commit()` first and only then do `rbac.AddProjectTreOwnerRole`?

internal/service/projects/main.go
```diff
@@ -562,6 +577,130 @@ func (s *Service) UpdateProjectTREDeployed(projectName string, data treopenapi.P
 	return nil
 }
 
+func (s *Service) ImportProjectTRE(data openapi.ProjectTREImport) error {
```
milanmlft, line 553: Add a call to `validateProjectTREAssetsAndMembers` here too?

internal/graceful/db.go
```diff
@@ -71,6 +72,10 @@ func InitDB() {
 // Create a new gorm DB, blocking until a connection is made
 func NewDB() *gorm.DB {
 	dbOnce.Do(func() {
```
milanmlft, line 74: from claude:

> dbOnce guards db inside NewDB(), but SetDBForTesting writes db directly without the mutex. When go test -p >1 runs multiple packages concurrently (e.g. projects and studies integration tests), both call SetDBForTesting and race on the package-level db. The race detector will flag this. Fix: use a sync.Mutex around db reads/writes, or reset dbOnce using sync.Once's documented workaround for tests.

Maybe not relevant for us

milanmlft (approved):
Looking good! Just minor comments

internal/service/projects/main.go
```diff
@@ -80,23 +83,31 @@ func (s *Service) validateProjectTREUpdate(data openapi.ProjectTREUpdate, projec
...
 
-	// Validate members
-	if len(members) > 0 {
-		if err := s.validateProjectMembers(members); err != nil {
-			return err
+	if err := s.validateProjectMembers(members); err != nil {
+		return err
+	}
+
+	minValidUid := 1001 // *MUST* not change
```
LakshithadeSilva: better to `const` this?

api/web.yaml
```diff
@@ -820,6 +820,26 @@ paths:
...
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/ProjectTREImport"
+      responses:
+        "204":
+          description: Project imported
+        "403":
```
LakshithadeSilva, line 835: Do we return a `401` as well since API has authn?

LakshithadeSilva (approved):
Nice! just a couple of weak comments.

internal/service/projects/main.go
t-young31: yes there is. however, the alternative: the tx commits but the rbac add fails is a worse state, because then there's a project which the owner can't access (rather than a project that doesn't exist that the owner could access if it did)

internal/service/projects/main.go
t-young31: not sure would want to block import of true state(?)

internal/graceful/db.go
t-young31: integration tests must be serial currently, so no additional issue (have #662 to improve)

api/web.yaml
t-young31: web auth is handled by the proxy, so this API won't directly return `401`s

internal/service/projects/main.go
t-young31: [c93567b](https://github.com/ucl-arc-tre/portal/pull/766/commits/c93567b3f94571d065d639ffadeb46a9a6f0b876)

PR #779, opened 2026-07-13
css tweaks

acholyn:
- consolidating some styles
- using variables for error colour, background and border colour
- renaming some shared styles for more flexible use

### DIFF
--- a/web/src/styles/form.css
+++ b/web/src/styles/form.css
@@ -27,6 +27,7 @@
         border-radius: var(--default-border-radius);
         font-size: 1rem;
         max-width: 400px;
+        width: clamp(200px, 30vw, 400px);
 
         &:focus {
           outline: none;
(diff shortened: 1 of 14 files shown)


web/src/styles/form.css
```diff
@@ -27,6 +27,7 @@
         border-radius: var(--default-border-radius);
         font-size: 1rem;
         max-width: 400px;
+        width: clamp(200px, 30vw, 400px);
```
t-young31: why? on narrow screen this will overflow(?)

acholyn: it _should_ choose the correct width for the screen, can rm the max above it actually
will double check 200px on mobile and maybe adjust it to be smaller

acholyn: the options with the numbers causes overflow on mobile

